PR #7, opened 2011-05-18
Bounded queue for a task executor

alf239:
temporary solution; it solves the current problem of the queue building up in the memory; it doesn't provide any significant advantage over the simple synchronised approach (well, the colours in profiler are nicer)

(cherry picked from commit 8bfb9b12475f7fee87f54b87f9648280f96b38a8)

Conflicts:
    atlas-index-api/src/main/java/uk/ac/ebi/gxa/statistics/ThreadSafeStatisticsBuilder.java


alf239:
still not too good; closing for a while

PR #8, opened 2011-05-20
Fixes for DB creation scripts

alf239:
In bries, the proper syntax for PKs is,

``` sql
USING INDEX TABLESPACE whatever
```

and for indexes it's just

``` sql
TABLESPACE whatever
```


alf239:
Oh, mea culpa — also it has an executors' quick fix

alf239:
@rpetry — please review

PR #9, opened 2011-05-23
Ticket #2961All samples in A2_Sample contains the same organismid

olgamelnichuk:
@alf239, @rpetry - could somebody please review this change. Thanks!


rpetry:
Reviewed and tested on my own schema

PR #13, opened 2011-06-06
Indexes made reasonably fast

alf239:
Indexes are reasonably fast now; more work is necessary, but it's not urgent any more.

The summary of changes is as follows:
- NetCDFs are read in one thread now: big NetCDFs require a lot of memory, so it's very hard for GC to handle more than one
- NetCDFs are sorted from big ones to smaller ones, so that we set up memory expectations early in the processing, and by the time the memory is filled with gene data, we only read small files
- min P / max T stats are surprisingly tough to handle; hence, we allocate a big thread pull (mostly idle) to handle those: that enables us to release file handles and large data arrays sooner, helping GC to do its job. Additionally, we don't wait for heavy map juggling and just proceed reading files.
- Experiment indexing is one-threaded now: it enables further Hibernate optimisations would we need those, and it does not affect the indexing time as the slowest part is gene indexing, not experiments'


atlas-index-api/src/main/java/uk/ac/ebi/gxa/statistics/Statistics.java
```diff
@@ -89,6 +89,45 @@ public class Statistics implements Serializable, StatisticsBuilder {
...
+    public void addAll(Statistics other) {
+        mergeMMC(statistics, other.statistics);
+        scoresAcrossAllEfos.addAll(other.scoresAcrossAllEfos);
+        mergeMMMC(pValuesTStatRanks, other.pValuesTStatRanks);
+        mergeMC(efAttributeToBioEntities, other.efAttributeToBioEntities);
+        mergeMC(efvAttributeToBioEntities, other.efvAttributeToBioEntities);
+        scoringExperiments.addAll(other.scoringExperiments);
+    }
+
+    private <T, V, D, M extends Map<V, Map<D, ConciseSet>>> void mergeMMMC(Map<T, M> a, Map<T, M> b) {
```
alf239, line 118: the weird "MMMC" name stands for "`merge`(`Map` of `Map` of `Map` of `ConciseSet`)". The only reason we need such a name is Java's type erasure: `merge(Map<T, M> a, Map<T, M> b)`, `merge(Map<T, Map<V, ConciseSet>> a, Map<T, Map<V, ConciseSet>> b)`, and `merge(Map<T, ConciseSet> a, Map<T, ConciseSet> b)`, as we'd all prefer to call these, would all be just `merge(Map, Map)`.

alf239:
Update:
- clarified tricky code, 
- fixed mistakes in progress messages
- switched to Hibernate 3.5.1-Final (mostly due to IntelliJ bug, but why not?)
- extracted some of the supplementary code so that both progress and processing logic are more clear now

alf239:
Merged with `develop`, ready for review. Anyone?

alf239:
Update:
- Instead of object indices, we're using real objects now: object reference takes the same 4 bytes as an `Integer` reference, so we just drop the indexes, and have code which is much more clear
- Eliminated duplicated strings from the serialized form 
- Eliminated `bioEntityIds` copying
- Made P and T primitives
- Adjusted T rank to be more similar to corresponding T values: we use these values instead of T, anyway
- Rearranged the methods to keep them closer to the corresponding data
- Made ExperimentInfo immutable and lean
- In total, won approximately 10% of disk space for the bitindex

atlas-index-api/src/main/java/uk/ac/ebi/gxa/statistics/EfoIndex.java
```diff
@@ -1,64 +1,59 @@
...
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
+import static com.google.common.collect.Maps.newHashMap;
+import static com.google.common.collect.Sets.newHashSet;
+
 /**
  * This class stores a mapping between efo terms and their corresponding ef-efv (i.e. Attribute)-Experiment index combinations
```
rpetry: Please update javadoc to reflect changes in data stractures used by this class

atlas-index-api/src/main/java/uk/ac/ebi/gxa/statistics/EfoIndex.java
```diff
@@ -1,64 +1,59 @@
...
 public class EfoIndex implements Serializable {
-
-    private static final long serialVersionUID = -1000094628265441595L;
+    private static final long serialVersionUID = 201106061646L;
 
     // efoTerm -> experiment index -> Set<Attribute Index>
-    private Map<String, Map<Integer, Set<Integer>>> efoIndex = new HashMap<String, Map<Integer, Set<Integer>>>();
+    private Map<String, Map<ExperimentInfo, Set<EfvAttribute>>> efoIndex = newHashMap();
 
     // Attribute -> experiment index -> efoTerm
```
rpetry: javadoc needs updating

atlas-index-api/src/main/java/uk/ac/ebi/gxa/statistics/ThreadSafeStatisticsBuilder.java
```diff
@@ -4,93 +4,72 @@ import com.google.common.collect.Multiset;
...
 
 @ThreadSafe
 public class ThreadSafeStatisticsBuilder implements StatisticsBuilder {
     private static final Logger log = LoggerFactory.getLogger(ThreadSafeStatisticsBuilder.class);
 
-    private final Statistics statistics = new Statistics();
-    private final ExecutorService executor = new ThreadPoolExecutor(1, 1, 1, TimeUnit.SECONDS,
-            new ArrayBlockingQueue<Runnable>(100, false), new BlockOnRejectedExecutionHandler());
+    @GuardedBy("threadLocals")
+    private final List<Statistics[]> threadLocals = new ArrayList<Statistics[]>();
```
rpetry: It would be really useful to see a javadoc on why we're having to use an array here

indexbuilder/src/main/java/uk/ac/ebi/gxa/index/builder/service/ExperimentAtlasIndexBuilderService.java
```diff
@@ -52,63 +51,33 @@ import static com.google.common.collect.Collections2.transform;
...
-                    if (firstError == null) {
-                        firstError = e;
-                    }
-                }
-            }
-
-            // if we have encountered an exception, throw the first error
-            if (firstError != null) {
-                throw new IndexBuilderException("An error occurred whilst building the Experiments index", firstError);
+            // fetch all experiments - check if we want all or only the pending ones
```
rpetry: Where do we check for pending experiments?

alf239: The check has long gone — thanks for spotting.

To be more precise, it has gone in revision 74e3d382 by @pashky:

> Ditch all index builder "update pending" logic as we're not going to support it in any foreseen future + and there go all  "pending genes" APIs as there ain't such thing
> 
> git-svn-id: svn://bar.ebi.ac.uk/branches/atlas-standalone@10971 2913f559-6b04-0410-9a09-c530ee9f5186

Dropped the comment as obsolete

alf239:
Well, indexing finishes perfectly for me. There are two other problems, though — not sure they are connected to this change, thus would like to make sure they aren't before proceeding.

alf239:
First problem, `ArrayIndexOutOfBoundsException` in E-GEOD-22001, is an existing one, and can be reproduced in `master`, hence not in scope of this change (tracked as Trac ticket 2995)

alf239:
Second one may or may not be introduced by the change: basically, when we draw an anatomogram (e.g. for Saa4/`ENSMUSG00000040017`), and one of the organism parts has an EFO (here `EFO_0000793`) having a child (here `EFO_0000844`) which is not present in the `EfoIndex`, we return `null` as a map from experiments to attributes, which obviously causes trouble.

The question stays open how come that we store that child in the first place.

atlas-index-api/src/main/java/uk/ac/ebi/gxa/statistics/ThreadSafeStatisticsBuilder.java
alf239: Extracted the trick in its own class, commented the latter

atlas-index-api/src/main/java/uk/ac/ebi/gxa/statistics/EfoIndex.java
alf239: done

atlas-index-api/src/main/java/uk/ac/ebi/gxa/statistics/EfoIndex.java
alf239: done

alf239:
Sorry, it was obvious: the children are taken from EFO, not from the index. Thus we cannot expect that index contains the EFO we have at hand. 

Fixed, anyway. 

@rpetry — please review.

PR #22, opened 2011-06-08
Double and float formatting, rounded to a proper number of digits

alf239:
A fix for `JsonResultRendererTest : No indent render (testNoIndentRender)`

@olgamelnichuk, could you please review as all others are busy?

### DIFF
--- a/atlas-utils/src/main/java/uk/ac/ebi/gxa/utils/FloatFormatter.java
+++ b/atlas-utils/src/main/java/uk/ac/ebi/gxa/utils/FloatFormatter.java
@@ -22,32 +22,12 @@
 
 package uk.ac.ebi.gxa.utils;
 
-import java.text.DecimalFormat;
+import static java.lang.Math.*;
 
 public abstract class FloatFormatter {
     private FloatFormatter() {
     }
 
-    private static DecimalFormat[] formatsCache = new DecimalFormat[10];
-
-    private static DecimalFormat decimalFormat(int numberOfDigits) {
-        DecimalFormat format = null;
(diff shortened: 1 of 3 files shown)


atlas-web/src/test/java/uk/ac/ebi/gxa/requesthandlers/base/restutil/JsonResultRendererTest.java
```diff
@@ -94,7 +94,7 @@ public class JsonResultRendererTest {
         r.render(o, sb, Object.class);
         Assert.assertEquals("Wrong format!", "atlas({\n" +
                 "    \"int\" : 123,\n" +
-                "    \"double\" : 1.23464E15,\n" +
+                "    \"double\" : 1.2346E15,\n" +
```
alf239, line 97: Note the number is rounded to the (supposedly) desired precision :)

atlas-utils/src/main/java/uk/ac/ebi/gxa/utils/FloatFormatter.java
```diff
@@ -55,9 +35,7 @@ public abstract class FloatFormatter {
         if (Float.isNaN(value)) {
             return "null";
```
alf239, line 36: For further review: do we really want to return `null` for `NaN`?

olgamelnichuk, line 36: The better way is to return "NaN" and "Infinity", but we already using null in javascript charts, so I suggest to postpone with this change for a while..

PR #25, opened 2011-06-21
Atlas 2.0.8 fixes: EBI menu fix

olgamelnichuk:
@alf239, @rpetry, please review this fix. I completely forgot about creating the trac issue, so I added all details below.

Here is the bug report from EBI external services:

There is a problem with http://www.ebi.ac.uk/gxa in Firefox and Chrome (Explorer fine) - I cannot access any links in the EBI header.
I think this is due to the positioning/padding given in atlas-ebi.css
.contents {
  position: relative !important;
  padding-top: 57px !important;
  top: 0 !important;
}
If you switch the 57 and 0, or use margin-top instead of padding-top the links become clickable, however not sure if this causes other problems elsewhere.

### DIFF
--- a/atlas-analytics/pom.xml
+++ b/atlas-analytics/pom.xml
@@ -26,11 +26,11 @@
     <parent>
         <artifactId>atlas</artifactId>
         <groupId>uk.ac.ebi.gxa</groupId>
-        <version>2.0.8</version>
+        <version>2.0.8.1-SNAPSHOT</version>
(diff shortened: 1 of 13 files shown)


alf239:
Please leave it as is. We should either do fixes in the `-fixes` branch or create merge requests _to_ this branch. You created the fix right in the branch — that should be okay for now. No need to merge it into `master`

PR #27, opened 2011-06-24
Wiggle fixes; sample <-> assay mapping

geometer:


alf239:
I believe I've already merged at least haf of those changes (e.g. getting rid of `NetCDFReader`) — is it really the commits you had in mind?

geometer:
You've merged wiggle fixes I think. This pull request is about sample <-> assay mappings.

alf239:
Well, it cannot be merged automatically, so it's not as if I could be of much help. Also, the diff doesn't seem to agree with the declared goal of the fix.

Could you please make sure that the requests are mergeable and in line with their description?

alf239:
Closing as obsolete — will be waiting for a new one

PR #31, opened 2011-06-24
Load data from local files

geometer:
Misha: please review this code


gxa:
Added a comment on the commit - would rather get rid of the boolean variable, and extract local processing into a method.

alf239:
@gxa @ostolop try choosing one identity :)

The comments you add to a commit in the pull request are automatically added to the pull request itself, there's no real need to comment twice (other than replies counter in the pull requests list)

ostolop:
:) Woops - wasn't sure about it.

PR #32, opened 2011-06-28
Trac2995 AIOOBE on E-GEOD-22001

olgamelnichuk:
For non symmetrical matrices of assay2sample mappings an OutOfBoundException were thrown.
Examples of experiments to verify the fix on: E-GEOD-22001, E-TABM-877

### DIFF
--- a/atlas-web/src/main/java/uk/ac/ebi/gxa/plot/AssayProperties.java
+++ b/atlas-web/src/main/java/uk/ac/ebi/gxa/plot/AssayProperties.java
@@ -78,7 +78,8 @@ public class AssayProperties {
             String[] vals = proxy.getCharacteristicValues(s);
             SampleCharacteristicsCompactData d = new SampleCharacteristicsCompactData(
                     nameConverter.apply(s), s2a[0].length);
-            for (int i = 0; i < vals.length; i++) {
+            assert vals.length == s2a.length;
+            for (int i = 0; i < s2a.length; i++) {
                 d.addScv(vals[i], i);
                 for (int j = s2a[i].length - 1; j >= 0; j--) {
                     if (s2a[i][j] > 0) {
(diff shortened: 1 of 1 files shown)


atlas-web/src/main/java/uk/ac/ebi/gxa/plot/AssayProperties.java
```diff
@@ -77,7 +77,7 @@ public class AssayProperties {
         for (String s : sampleCharacteristics) {
             String[] vals = proxy.getCharacteristicValues(s);
             SampleCharacteristicsCompactData d = new SampleCharacteristicsCompactData(
-                    nameConverter.apply(s), vals.length);
+                    nameConverter.apply(s), s2a[0].length);
             for (int i = 0; i < vals.length; i++) {
```
alf239: While I do know that `vals.length == s2a.length`, it would be much more clear for the programmer to see simple iteration over 2D matrix, e.g. `for (int i = 0; i < s2a.length; i++)`

alf239: `assert vals.length == s2a.length;` would be nice as well (while admittedly useless in current setup)

atlas-web/src/main/java/uk/ac/ebi/gxa/plot/AssayProperties.java
```diff
@@ -77,8 +77,9 @@ public class AssayProperties {
         for (String s : sampleCharacteristics) {
             String[] vals = proxy.getCharacteristicValues(s);
             SampleCharacteristicsCompactData d = new SampleCharacteristicsCompactData(
-                    nameConverter.apply(s), vals.length);
-            for (int i = 0; i < vals.length; i++) {
+                    nameConverter.apply(s), s2a[0].length);
```
alf239, line 80: OK for hotfix, please rephrase in `develop`

PR #40, opened 2011-06-30
Trac3006 Display EF for EFVs in autocomplete

olgamelnichuk:
Now showing all found efv in the autocomplete. Factor name added to the view of efv autocomplte item.

### DIFF
--- a/atlas-web/src/main/java/ae3/service/structuredquery/AtlasEfvService.java
+++ b/atlas-web/src/main/java/ae3/service/structuredquery/AtlasEfvService.java
@@ -41,6 +41,7 @@ import uk.ac.ebi.gxa.statistics.StatisticsType;
 import uk.ac.ebi.microarray.atlas.model.Property;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import java.util.*;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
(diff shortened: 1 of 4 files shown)


alf239:
Seems fine; the only question is, we have `biopsy tissue` twice here — how come?

screencap

olgamelnichuk:
There are two `biopsy tissue` : `liver`, one with a space at the end and another without space. To check this, go to the production database and run the following query:

``` sql
SELECT min(p.propertyid), p.name, min(pv.propertyvalueid), pv.name 
  FROM a2_property p, a2_propertyvalue pv 
 WHERE  pv.propertyid=p.propertyid AND p.name='biopsy tissue' 
 and pv.name like 'liver%' GROUP BY p.name, pv.name
```

alf239:
@ostolop — the change itself is okay, are we okay with semi-duplicated EFVs?

ostolop:
disease_state and organism_part should come from curatedEF

olgamelnichuk:
Curated factor names should be shown now

olgamelnichuk:
Earlier it was just a quick solution. Property is not exactly what we want to show here. The default view of autocomplete item doesn't require to show its property in gray color after its name. So I prefer to show exactly what is required. If there is no factorName then show nothing.

PR #42, opened 2011-07-01
Install scripts + NetCDF update fix

alf239:
The scripts making it possible to deploy a server in one command. Quite naïve, but they do work (tested on `lime` and staging).

Along with the scripts comes a fix I needed to apply: NetCDF update must be enclosed into a Hibernate/DAO session, which was missed before. Fixed now, the change is trivial.

Please review


alf239:
As a side effect of the change, old `nightly` profile became obsolete, and so did the variables in `~/.m2/settings.xml`

alf239:
During tests on the `NIGHTLY` build, added two minor fixes:
- ignoring unknown NetCDFs (it's true we'd better die so that the problem could be fixed, but imagine a end-eser trying to get the app up)
- fixed an NPE in the anatomogram code — I believe it's a regression, I did fix it already

alf239:
@ostolop, @rpetry, @olgamelnichuk, anyone — please review when you have a minute :)

PR #44, opened 2011-07-04
Ticket #3003 Autocomplete clipping long text

olgamelnichuk:
@alf239, @rpetry - please, review


alf239:
We have absolutely the same problem with genes (e.g. if you enter `i`, you'll get _I-2_ gene with _Drosophila melanogaster_ clipped on the right)

alf239:
For `I-2`, `&nbsp;` gets escaped in the pop up. I suggest we stop and discuss what exactly we're trying to achieve — it seems we've got a broken telephone effect here.

PR #47, opened 2011-07-05
Ticket #2980 - shortend (ellipsis) the user's query in the downloads pop-up when it is longer than 50 chars - to allow the user to scroll to the rightmost column containing the link to the downloaded data...

rpetry:
alf, olkin - could you review please?

### DIFF
--- a/atlas-web/src/main/webapp/WEB-INF/jsp/query/downloads.jsp
+++ b/atlas-web/src/main/webapp/WEB-INF/jsp/query/downloads.jsp
@@ -89,9 +89,17 @@
                         <th class="padded header">Download Progress</th>
                         <th class="padded header">File</th>
                         <c:forEach items="${downloads}" var="download" varStatus="i">
+                          <c:choose>
+                            <c:when test="${f:length(download.value.query) > 50}">
+                                <c:set var="queryString" value="${f:substring(download.value.query,0,50)}..." />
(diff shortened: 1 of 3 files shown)


atlas-web/src/main/webapp/WEB-INF/jsp/query/downloads.jsp
```diff
@@ -89,9 +89,17 @@
...
                         <c:forEach items="${downloads}" var="download" varStatus="i">
+                          <c:choose>
+                            <c:when test="${f:length(download.value.query) > 50}">
+                                <c:set var="queryString" value="${f:substring(download.value.query,0,50)}..." />
+                            </c:when>
+                            <c:otherwise>
+                                <c:set var="queryString" value="${download.value.query}" />
+                            </c:otherwise>
+                          </c:choose>
                             <tr>
```
olgamelnichuk, line 100: Just a minor note. If the query length is 51, than we clip it to 50 and add 3 more ("..."), so the final length is 53. It is a bit unfair because we can fit the original string in 53 characters :)

So, the first 50 should be changed to 53, or second 50 should be changed on 47.

rpetry, line 100: good point - done.

PR #49, opened 2011-07-06
Privacy buttons

alf239:
Buttons allowing admin to make experiment public and/or private; also, the API doing the same.

The code is awful, but is left as such until further changes in the `develop` branch.

### DIFF
--- a/atlas-web/src/main/webapp/admin.html
+++ b/atlas-web/src/main/webapp/admin.html
@@ -170,10 +170,12 @@
                 <input type="button" value="Repair" class="repair">
                 <input type="button" value="Unload" class="unload">
                 |
+                <input type="button" value="Make private" class="makeprivate">
+                <input type="button" value="Make public" class="makepublic">
+                |
                 <input type="button" value="Recompute analytics" class="analytics">
                 <input type="button" value="Update NetCDFs" class="updatenetcdf">
                 <input type="button" value="Update Index" class="updateindex">
-                <input type="button" value="Release" class="datarelease">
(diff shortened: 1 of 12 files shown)


atlas-web/src/main/webapp/admin.html
```diff
@@ -170,10 +170,12 @@
                 <input type="button" value="Repair" class="repair">
                 <input type="button" value="Unload" class="unload">
                 |
+                <input type="button" value="Make private" class="makeprivate">
+                <input type="button" value="Make public" class="makepublic">
+                |
                 <input type="button" value="Recompute analytics" class="analytics">
                 <input type="button" value="Update NetCDFs" class="updatenetcdf">
                 <input type="button" value="Update Index" class="updateindex">
-                <input type="button" value="Release" class="datarelease">
```
alf239, line 176: Release doesn't really work, thus hidden. My apologies for making it in this particular patch.

scripts/deploy/context.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<Context swallowOutput="true" docBase="${WAR_FILE}" unpackWAR="true" antiJARLocking="true" antiResourceLocking="true">
+<Context swallowOutput="true" docBase="${WAR_FILE}" unpackWAR="false">
```
alf239, line 2: Pecularities of the update process: tomcat tends to leave old HTMLs intact on redeploy, this combination of settings lets us avoid the mess

PR #52, opened 2011-07-08
Deploy scripts are extracted into a separate repository

alf239:
Deploy scripts were extracted into a separate repository in order to prevent accidental credentials leaks


alf239:
The context example can be found at https://github.com/gxa/gxa/wiki/Installation, so removal does not do any harm

PR #53, opened 2011-07-18
Curation api - changes by @ostolop and @rpetry

rpetry:
Could someone please review?


atlas-web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -473,7 +482,7 @@
     </filter-mapping>
 
     <filter-mapping>
-        <filter-name>OpenSessionInViewFilter</filter-name>
+        <filter-name>ReadWriteHibernateSessionFilter</filter-name>
```
alf239: We replaced the filter mapping, but did not remove the older filter. Is it as intended?

atlas-dao/src/main/resources/uk/ac/ebi/gxa/dao/daoContext.xml
```diff
@@ -1,7 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <beans xmlns="http://www.springframework.org/schema/beans"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-       xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd">
+       xmlns:tx="http://www.springframework.org/schema/tx"
+       xmlns:aop="http://www.springframework.org/schema/aop"
```
alf239: AOP? Why?

alf239:
Does not compile.

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/api/ApiArrayDesign.java
```diff
@@ -0,0 +1,66 @@
...
+ * @author Misha Kapushesky
+ */
+public class ApiArrayDesign {
+    private String accession;
+    private String name;
+    private String provider;
+    private String type;
+//    private Long mappingSoftwareId;
+//    private Map<String, Long> designElements = new HashMap<String, Long>();
+//    private Map<Long, List<Long>> genes = new HashMap<Long, List<Long>>();
```
alf239: Could we please get rid of commented-out code? No one will ever be able to find out why it was left here.

atlas-dao/src/main/java/uk/ac/ebi/gxa/dao/SampleDAO.java
```diff
@@ -0,0 +1,25 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.ac.ebi.microarray.atlas.model.Sample;
+
+/**
+ * Created by IntelliJ IDEA.
+ * User: rpetry
+ * Date: 7/13/11
+ * Time: 3:57 PM
+ * To change this template use File | Settings | File Templates.
```
alf239: `use File | Settings | File Templates.`

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/Assay.java
```diff
@@ -214,6 +214,46 @@ public class Assay {
...
+    }
+
+    public AssayProperty getProperty(PropertyValue propertyValue) {
+        for (AssayProperty property : properties) {
+            if (property.getPropertyValue().equals(propertyValue))
+                return property;
+        }
+
+        return null;
+    }
```
alf239, line 224: The three methods above are suspiciously similar :)

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/Sample.java
```diff
@@ -106,6 +103,8 @@ public class Sample {
         return channel;
     }
 
+
+
     public long getSampleID() {
```
alf239: it must be `Long` I believe

atlas-utils/src/main/java/uk/ac/ebi/gxa/utils/GuavaUtil.java
```diff
@@ -0,0 +1,37 @@
...
+import uk.ac.ebi.gxa.exceptions.LogUtil;
+
+import javax.annotation.Nullable;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+
+/**
+ * @author Misha Kapushesky
+ */
+public class GuavaUtil {
```
alf239: It has nothing to do with Guava actually. `CollectionUtil` perhaps?

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,513 @@
...
+     */
+    @Transactional
+    public void saveExperiment(@Nonnull final ApiExperiment apiExperiment, final HttpServletResponse response) {
+        Experiment experiment = atlasDAO.getExperimentByAccession(apiExperiment.getAccession());
+
+        if (experiment != null) {
+            // TODO: 4ostolop: should we or should we not keep it?
+
+//            log.info("Deleting experiment " + experiment.getAccession() + " in order to update");
+//            experimentDAO.delete(experiment);
```
alf239: should we or should we not keep it?

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,513 @@
...
+    /**
+     * Adds (or updates mapping to efo terms for) assayProperties to assay: assayAccession in experiment: experimentAccession
+     *
+     * @param experimentAccession
+     * @param assayAccession
+     * @param assayProperties
+     * @param response
+     * @throws ResourceNotFoundException if experiment: experimentAccession or assay: assayAccession in that experiment are not found
+     */
+    @Transactional
```
alf239, line 262: It's enough to make the whole class `@Transactional`, I think. Do we really need non-transactional methods here?

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,470 @@
...
+     * If entity == null, this method sets appropriate response status and then throws ResourceNotFoundException
+     *
+     * @param entity
+     * @param accession
+     * @param <T>
+     * @throws ResourceNotFoundException
+     */
+    private <T> void checkIfFound(T entity, String accession) throws ResourceNotFoundException {
+        if (entity == null) {
+            throw new ResourceNotFoundException("No records for " + entity.getClass().getName() + accession);
```
alf239: # Null pointer dereference

A `null` pointer is dereferenced here.  This will lead to a `NullPointerException` when the code is executed.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,470 @@
...
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+import uk.ac.ebi.gxa.dao.*;
+import uk.ac.ebi.gxa.web.controller.ResourceNotFoundException;
```
alf239: It's still `web.controller`, we cannot use it if we want to keep code clean.

alf239:
@olgamelnichuk, could you please review the Controller part? I'm not good at mapping exceptions to HTTP statuses; e.g. should I update status by hand if I throw an exception?

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/api/ApiArrayDesign.java
rpetry: Done

atlas-dao/src/main/java/uk/ac/ebi/gxa/dao/SampleDAO.java
rpetry: Ammended

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/Assay.java
rpetry, line 224: Refactored

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/Sample.java
rpetry: Corrected - also in Assay class

atlas-dao/src/test/java/uk/ac/ebi/gxa/dao/AtlasDAOTestCase.java
```diff
@@ -404,18 +402,18 @@ public abstract class AtlasDAOTestCase extends DataSourceBasedDBTestCase {
...
-        runStatement(conn, "CREATE SEQUENCE A2_ORGANISM_SEQ");
-        runStatement(conn, "CREATE SEQUENCE A2_PROPERTY_SEQ");
-        runStatement(conn, "CREATE SEQUENCE A2_PROPERTYVALUE_SEQ");
-        runStatement(conn, "CREATE SEQUENCE A2_SAMPLE_SEQ");
-        runStatement(conn, "CREATE SEQUENCE A2_SAMPLEPV_SEQ");
+        runStatement(conn, "CREATE SEQUENCE A2_ARRAYDESIGN_SEQ START WITH 10000000");
+        runStatement(conn, "CREATE SEQUENCE A2_ASSAY_SEQ START WITH 10000000");
+        runStatement(conn, "CREATE SEQUENCE A2_ASSAYPV_SEQ START WITH 10000000");
+        runStatement(conn, "CREATE SEQUENCE A2_ASSET_SEQ START WITH 10000000");
+        runStatement(conn, "CREATE SEQUENCE A2_EXPERIMENT_SEQ START WITH 10000000");
```
alf239, line 427: Actually it doesn't work for the experiments: they're using "proper" IDs (as we needed to have matching NetCDF name long, long ago), and real values are around 10E7. We should update DB snapshot, I believe.

atlas-dao/src/test/java/uk/ac/ebi/gxa/dao/AtlasDAOTestCase.java
```diff
@@ -353,14 +353,12 @@ public abstract class AtlasDAOTestCase extends DataSourceBasedDBTestCase {
...
         runStatement(conn,
                 "  CREATE TABLE A2_ASSAYPVONTOLOGY (\n" +
-                        "    ASSAYPVONTOLOGYID bigint not null\n" +
-                        "  , ONTOLOGYTERMID bigint not null\n" +
+                        "    ONTOLOGYTERMID bigint not null\n" +
                         "  , ASSAYPVID bigint not null)");
 
         runStatement(conn,
                 "  CREATE TABLE A2_SAMPLEPVONTOLOGY (\n" +
-                        "    SAMPLEPVONTOLOGYID bigint not null\n" +
```
alf239, line 362: We need to port the same change to the "real" database as well There's no point adding IDs to an association table. Even if triggers save us from this problem, it's a bad, bad design.

atlas-dao/src/test/java/uk/ac/ebi/gxa/dao/AtlasDAOTestCase.java
```diff
@@ -353,14 +353,12 @@ public abstract class AtlasDAOTestCase extends DataSourceBasedDBTestCase {
 
         runStatement(conn,
                 "  CREATE TABLE A2_ASSAYPVONTOLOGY (\n" +
-                        "    ASSAYPVONTOLOGYID bigint not null\n" +
```
alf239, line 356: We need to port the same change to the "real" database as well There's no point adding IDs to an association table. Even if triggers save us from this problem, it's a bad, bad design.

atlas-utils/src/main/java/uk/ac/ebi/gxa/utils/GuavaUtil.java
rpetry: Renamed to TransformerUtil, as agreed

atlas-dao/src/test/java/uk/ac/ebi/gxa/dao/TestAtlasDAO.java
```diff
@@ -178,4 +206,30 @@ public class TestAtlasDAO extends AtlasDAOTestCase {
...
+        final Assay assay = experiment.getAssay(ABC_ABCXYZ_SOME_THING_1234_ABC123);
+        final PropertyValue propertyValue = atlasDAO.getOrCreatePropertyValue(PROPERTY_NAME, PROPERTY_VALUE);
+        List<OntologyTerm> terms = Collections.singletonList(atlasDAO.getOrCreateOntologyTerm(
+                ONTOLOGY_TERM,
+                null,
+                null,
+                ONTOLOGY_NAME,
+                ONTOLOGY_DESCRIPTION,
+                null,
+                ONTOLOGY_VERSION));
```
alf239: This line reminds me that most of the `Ontology` and `OntologyTerm` fields are unused. Actually, we only need term's accession. We can of course track ontologies as well, but we don't.

Out of scope, anyway. Created a ticket, http://bar.ebi.ac.uk:8080/trac/ticket/3058

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry, line 262: Keeping all service setter/getter methods in CurationService keeps separation of concerns clear - the service deals with DAO access, the controller looks after HTTP transport. Assuming we agree on the above, annotating @Transactional at the class level would make the getter calls more costly than they need to.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry: Good spot - fixed.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry: Moved out ResourceNotFoundException to atlas-utls

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
alf239, line 262: Well, I see no connection between separation of concerns and positioning of `@Transactional`. 

As of the costs, you may be right, I'll check.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,469 @@
...
+                apiOntologyTerm.getTerm(),
+                apiOntologyTerm.getDescription(),
+                apiOntologyTerm.getOntology().getName(),
+                apiOntologyTerm.getOntology().getDescription(),
+                apiOntologyTerm.getOntology().getSourceUri(),
+                apiOntologyTerm.getOntology().getVersion());
+    }
+
+    /**
+     * If entity == null, this method sets appropriate response status and then throws ResourceNotFoundException
```
alf239: There's no `appropriate response status` any more. Moreover, checking for `null` is a bad practice: if we have an accessin, we're almost sure we know what we're asking for. Not having such an object is an exception.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,469 @@
...
+                List<OntologyTerm> terms = Lists.newArrayList();
+                for (ApiOntologyTerm apiOntologyTerm : apiSampleProperty.getTerms()) {
+                    terms.add(atlasDAO.getOrCreateOntologyTerm(
+                            apiOntologyTerm.getAccession(),
+                            apiOntologyTerm.getTerm(),
+                            apiOntologyTerm.getDescription(),
+                            apiOntologyTerm.getOntology().getName(),
+                            apiOntologyTerm.getOntology().getDescription(),
+                            apiOntologyTerm.getOntology().getSourceUri(),
+                            apiOntologyTerm.getOntology().getVersion()));
```
alf239: We have the same code repeated 30 lines above. Can't we just create a method?

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,469 @@
...
+     */
+    private OntologyTerm getOrCreateOntologyTerm(@Nonnull ApiOntologyTerm apiOntologyTerm) {
+        return atlasDAO.getOrCreateOntologyTerm(
+                apiOntologyTerm.getAccession(),
+                apiOntologyTerm.getTerm(),
+                apiOntologyTerm.getDescription(),
+                apiOntologyTerm.getOntology().getName(),
+                apiOntologyTerm.getOntology().getDescription(),
+                apiOntologyTerm.getOntology().getSourceUri(),
+                apiOntologyTerm.getOntology().getVersion());
```
alf239: We effectively duplicate the above method here, which isn't a good idea. Also, please consider the [Law of Demeter](http://en.wikipedia.org/wiki/Law_of_Demeter): taking `apiOntologyTerm.getOntology()` apart is akin to commanding dog's legs.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
alf239: On a second though, we already have a method, `private OntologyTerm getOrCreateOntologyTerm(@Nonnull ApiOntologyTerm apiOntologyTerm)`. starting line 411. Get rid of unnecessary code duplication please.

atlas-dao/pom.xml
```diff
@@ -94,6 +94,11 @@
             <artifactId>spring-core</artifactId>
             <version>${spring.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-web</artifactId>
+            <version>${spring.version}</version>
+        </dependency>
```
olgamelnichuk, line 105: Why do we need spring-web in atlas-dao module?

atlas-web/src/main/java/uk/ac/ebi/gxa/web/controller/api/CurationApiController.java
```diff
@@ -0,0 +1,220 @@
...
+    @Autowired
+    private CurationService curationService;
+
+    @RequestMapping(value = "/experiments/{experimentAccession}",
+            method = RequestMethod.GET)
+    public ApiExperiment getExperiment(@PathVariable("experimentAccession") final String experimentAccession,
+                                       HttpServletResponse response) throws ResourceNotFoundException {
+        try {
+            ApiExperiment apiExperiment = curationService.getExperiment(experimentAccession);
+            response.setStatus(HttpServletResponse.SC_FOUND);
```
olgamelnichuk: According to the javadoc,

``` java
    /**
     * Status code (302) indicating that the resource reside
     * temporarily under a different URI. Since the redirection might
     * be altered on occasion, the client should continue to use the
     * Request-URI for future requests.(HTTP/1.1) To represent the
     * status code (302), it is recommended to use this variable.
     */
    public static final int SC_FOUND = 302;
```

Why do we set this status here?

atlas-web/src/main/java/uk/ac/ebi/gxa/web/controller/api/CurationApiController.java
```diff
@@ -0,0 +1,220 @@
...
+            return apiExperiment;
+        } catch (ResourceNotFoundException e) {
+            response.setStatus(HttpServletResponse.SC_NOT_FOUND);
```
olgamelnichuk: We already have exception handler in the AtlasViewController:

``` java
    @ExceptionHandler(ResourceNotFoundException.class)
    @ResponseStatus(HttpStatus.NOT_FOUND)
    public ModelAndView handleError(ResourceNotFoundException e) {
        log.info(e.getMessage());
        return errorDetails(e);
    }
```

atlas-web/src/main/java/uk/ac/ebi/gxa/web/controller/api/CurationApiController.java
```diff
@@ -0,0 +1,220 @@
...
+            throw e;
+        }
+    }
+
+    @RequestMapping(value = "/experiments/{experimentAccession}",
+            method = RequestMethod.PUT)
+    public void putExperiment(@RequestBody final ApiExperiment apiExperiment,
+                              HttpServletResponse response) throws ResourceNotFoundException {
+        curationService.saveExperiment(apiExperiment);
+        response.setStatus(HttpServletResponse.SC_CREATED);
```
olgamelnichuk: Please use spring annotation to set the response status:

``` java
@ResponseStatus(HttpStatus.CREATED)
```

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry: Removed the outdated comment. Not sure I understand the bad practice bit - having null here means that we could not find the object used specified, not that we didn't have the object in the first place. The user may know what object they're looking for, but but may mistype it.

atlas-dao/src/test/java/uk/ac/ebi/gxa/dao/AtlasDAOTestCase.java
alf239, line 362: Addressed in e32febb60c87f22cb52bf3f79e262da8bd5a0854; will be cherry-picked to develop, too.

atlas-dao/src/test/java/uk/ac/ebi/gxa/dao/AtlasDAOTestCase.java
alf239, line 356: Addressed in e32febb; will be cherry-picked to develop, too.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry: Done.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry: Modified, as we discussed, to force the user to create a new ontology before assigning ontology terms to it.

atlas-dao/pom.xml
rpetry, line 105: Have removed - must have been a remnant of some earlier attempts - thanks for spotting it.

atlas-web/src/main/java/uk/ac/ebi/gxa/web/controller/api/CurationApiController.java
rpetry: Replaced it with SC_OK - am sure that's what Misha intended to use.

atlas-web/src/main/java/uk/ac/ebi/gxa/web/controller/api/CurationApiController.java
rpetry: Thanks—wasn't aware of that—have removed exception handling from `CurationApiController`.

atlas-web/src/main/java/uk/ac/ebi/gxa/web/controller/api/CurationApiController.java
rpetry: Done.

atlas-web/src/main/java/uk/ac/ebi/gxa/web/controller/api/CurationApiController.java
olgamelnichuk: Actually SC_OK is set by default for all responses finished successfully

olgamelnichuk:
I don't see any unit tests for db service and trasformation utilities.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry: As discussed, removed `saveExperiment()` method for now. If `experimentDAO.delete(experiment)` is left commented out, this method will create a new experiment with the same accession but different `experimentid` (wrong). If `experimentDAO.delete(experiment)` is uncommented, the old experiment with same accession willl be replaced with the new one, but also with a different `experimentid`. This is kind of what experiment re-loading currently does, but the new experimentid in the save experiment means that all the processing history for that experiment's accession in Atlas admin API is lost, which is not desirable either.

The 'save experiment' functionality is not currently offered as part of the Curation API. Until the requirements on saving experiments are fully clarified, @alf239 and I made a decision to remove `saveExperiment()` method.

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/ArrayDesign.java
```diff
@@ -36,13 +36,13 @@ public class ArrayDesign {
...
     @SequenceGenerator(name = "arrayDesignSeq", sequenceName = "A2_ARRAYDESIGN_SEQ")
-    private long arrayDesignID;
+    private Long arrayDesignID;
     private String accession;
     private String name;
     private String provider;
     private String type;
     @Column(name = "MAPPINGSWID")
-    private long mappingSoftwareId;
+    private Long mappingSoftwareId;
```
alf239, line 45: The getter will throw a hard-to-diagnose NPE if called when `mappingSoftwareId  == null`

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/ArrayDesign.java
```diff
@@ -36,13 +36,13 @@ public class ArrayDesign {
     @Id
     @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "arrayDesignSeq")
     @SequenceGenerator(name = "arrayDesignSeq", sequenceName = "A2_ARRAYDESIGN_SEQ")
-    private long arrayDesignID;
+    private Long arrayDesignID;
```
alf239, line 39: The getter will throw a hard-to-diagnose NPE if called when `arrayDesignID == null`

sql/Schema/Tables.sql
```diff
@@ -824,14 +824,13 @@ ALTER TRIGGER A2_SAMPLE_INSERT ENABLE;
 
 
  CREATE TABLE "A2_ASSAYSAMPLE" (
-    "ASSAYSAMPLEID" NUMBER(22,0) CONSTRAINT NN_ASSAYSAMPLE_ID NOT NULL
-  , "ASSAYID" NUMBER(22,0) CONSTRAINT NN_ASSAYSAMPLE_ASSAYID NOT NULL
+   "ASSAYID" NUMBER(22,0) CONSTRAINT NN_ASSAYSAMPLE_ASSAYID NOT NULL
   , "SAMPLEID" NUMBER(22,0) CONSTRAINT NN_ASSAYSAMPLE_SAMPLEID NOT NULL) ;
```
alf239, line 822: Since we never rebuild the curators' instance from scratch, an update script would be helpful. Please refer to https://github.com/gxa/gxa/blob/develop/sql/updates/2.0.8-2.0.9/20110811-1-proper_associations.sql for an example.

sql/Schema/Tables.sql
```diff
@@ -861,26 +860,6 @@ ALTER TRIGGER A2_SAMPLE_INSERT ENABLE;
...
-    UNIQUE("ASSAYID","SAMPLEID") 
-    ENABLE;  
-
-
-CREATE OR REPLACE TRIGGER A2_ASSAYSAMPLE_INSERT  
-before insert on A2_AssaySample
-for each row
-begin
-if(:new.AssaySampleID is null) then
-select A2_AssaySample_seq.nextval into :new.AssaySampleID from dual;
```
alf239, line 876: We're removing the usage of `A2_AssaySample_seq`, bur not `A2_AssaySample_seq` itself, why?

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,538 @@
...
+     * @throws ResourceNotFoundException
+     */
+    @Transactional
+    public void removePropertyValue(final String propertyName,
+                                    final String propertyValue) throws ResourceNotFoundException {
+
+        Property property = propertyDAO.getByName(propertyName);
+        checkIfFound(property, Property.class, propertyName);
+        PropertyValue propValue = propertyValueDAO.find(property, propertyValue);
+        checkIfFound(propValue, PropertyValue.class, propertyValue);
```
alf239: Use checked exceptions instead, please. You will throw an exception anyway, it's just that this exceptional situation pollutes the code: we return `null` in DAO, making check for that, and then we're trying not to forget this `null` check in the clients. Exceptions are technically impossible to forget, and they structure the code better.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,538 @@
...
+        List<Assay> assays = assayDAO.getAssaysByPropertyValue(propertyValue);
+        for (Assay assay : assays) {
+            assay.deleteProperty(propValue);
+            assayDAO.save(assay);
+        }
+        List<Sample> samples = sampleDAO.getSamplesByPropertyValue(propertyValue);
+        for (Sample sample : samples) {
+            sample.deleteProperty(propValue);
+            sampleDAO.save(sample);
+        }
```
alf239, line 119: Can't we just enable cascading on delete?

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,538 @@
...
+
+        // Now delete propertyName:propertyValue itself
+        propertyDAO.delete(property, propValue);
+        propertyDAO.save(property);
```
alf239: Two calls to one object in a row almost always mean you're doing something that object would've pretty much able to do on its own.

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/ArrayDesign.java
rpetry, line 45: Made getMappingSoftwareId() return Long

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/ArrayDesign.java
rpetry, line 39: Made getArrayDesignID() return Long

sql/Schema/Tables.sql
rpetry, line 876: You're right - we should - I missed that due to different casing in the creation statement - I've removed it now.

sql/Schema/Tables.sql
rpetry, line 822: Added DB update files into 2.0.9-2.0.10 dir.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry, line 119: Sorry - missed that - cascading was already enabled - I didn't need to cycle through assays/samples here - thanks

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry: Yep - moved the second call to  propertyDAO.delete()

olgamelnichuk:
Why we start to use microarray/atlas in the package name (e.g. uk/ac/ebi/microarray/atlas/api/)?

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/dao/LoaderDAO.java
```diff
@@ -36,11 +43,22 @@ public class LoaderDAO {
...
-        PropertyValue propertyValue = propertyValueDAO.find(property, value);
+        PropertyValue propertyValue = null;
+
+        try {
+            propertyValue = propertyValueDAO.find(property, value);
+        } catch (DAOException e) {
+            // do nothing - valid situation
+        }
         if (propertyValue == null) {
             propertyValueDAO.save(propertyValue = new PropertyValue(null, property, value));
```
alf239: Why cannot we do it "catch" clause and keep login in one place?

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/dao/LoaderDAO.java
```diff
@@ -27,7 +28,13 @@ public class LoaderDAO {
...
-        Organism organism = organismDAO.getByName(name);
+        Organism organism = null;
+        try {
+            organism = organismDAO.getByName(name);
+        } catch (DAOException e) {
+            // do nothing - valid situation
+        }
+
         if (organism == null) {
             organismDAO.save(organism = new Organism(null, name));
```
alf239: Why cannot we do it "catch" clause and keep login in one place?

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,533 @@
...
+
+        Ontology ontology = null;
+        try {
+            ontology = atlasDAO.getOntologyByName(apiOntology.getName());
+        } catch (DAOException e) {
+            // Do nothing - valid situation
+        }
+
+        if (ontology == null) {
+            ontology = getOrCreateOntology(apiOntology);
```
alf239: Why cannot we do it "catch" clause and keep login in one place?

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,533 @@
...
+        } else {
+            ontology.setDescription(apiOntology.getDescription());
+            ontology.setName(apiOntology.getName());
+            ontology.setVersion(apiOntology.getVersion());
+            ontology.setSourceUri(apiOntology.getSourceUri());
```
alf239, line 400: As far as I can see, no test cover this branch. Do we really need it?

alf239:
As far as I can see, all the `// do nothing - valid situation` situations can and should be handled in the `catch` clause, there's no need to check for `null` after `try`...`catch`.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,533 @@
...
+            }
+
+            if (ontologyTerm == null) {
+                ontologyTerm = getOrCreateOntologyTerm(apiOntologyTerm);
+            } else {
+                ontologyTerm.setAccession(apiOntologyTerm.getAccession());
+                ontologyTerm.setDescription(apiOntologyTerm.getDescription());
+                Ontology ontology = getOrCreateOntology(apiOntologyTerm.getOntology());
+                ontologyTerm.setOntology(ontology);
+                ontologyTerm.setTerm(apiOntologyTerm.getTerm());
```
alf239, line 435: This branch is not covered; do we need it?

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,533 @@
...
+            PropertyValue newPropertyValue = atlasDAO.getOrCreatePropertyValue(propertyName, newValue);
+
+            List<Sample> samples = sampleDAO.getSamplesByPropertyValue(oldValue);
+            for (Sample sample : samples) {
+                SampleProperty oldSampleProperty = sample.getProperty(oldPropertyValue);
+                SampleProperty newSampleProperty = sample.getProperty(newPropertyValue);
+                List<OntologyTerm> terms = oldSampleProperty.getTerms();
+                sample.deleteProperty(oldPropertyValue);
+                if (newSampleProperty != null) {
+                    terms.addAll(newSampleProperty.getTerms());
```
alf239, line 188: This branch is not covered

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,533 @@
...
+        Assay assay = findAssay(experimentAccession, assayAccession);
+
+        for (ApiProperty apiAssayProperty : assayProperties) {
+            PropertyValue propertyValue = atlasDAO.getOrCreatePropertyValue(
+                    apiAssayProperty.getPropertyValue().getProperty().getName(),
+                    apiAssayProperty.getPropertyValue().getValue());
+
+            List<OntologyTerm> terms = Lists.newArrayList();
+            for (ApiOntologyTerm apiOntologyTerm : apiAssayProperty.getTerms()) {
+                terms.add(getOrCreateOntologyTerm(apiOntologyTerm));
```
alf239, line 273: This cycle is never tested, as `apiAssayProperty.getTerms()` is always empty

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,532 @@
...
+import static com.google.common.collect.Collections2.transform;
+
+/**
+ * This class handles all Curation API requests, delegated from CurationApiController
+ *
+ * @author Misha Kapushesky
+ */
+@Service
+public class CurationService {
+    private final Logger log = LoggerFactory.getLogger(getClass());
```
alf239: We never log anything here

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/dao/LoaderDAO.java
rpetry: I guess we can - it makes the code simpler. done.

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/dao/LoaderDAO.java
rpetry: Done

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry: Done

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry, line 400: testPutOntology() in TestCurationService, no?

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry, line 435: testPutOntologyTerms() in TestCurationService does

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry, line 188: See testReplacePropertyValueInSamples() in TestCurationService

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,526 @@
...
+        for (ApiOntologyTerm apiOntologyTerm : apiOntologyTerms) {
+            OntologyTerm ontologyTerm;
+            try {
+                ontologyTerm = atlasDAO.getOntologyTermByAccession(apiOntologyTerm.getAccession());
+            } catch (DAOException e) { // ontology term not found - create a new one
+                ontologyTerm = getOrCreateOntologyTerm(apiOntologyTerm);
+            }
+
+            if (ontologyTerm == null) { // ontology term already exists - update it
+                ontologyTerm.setAccession(apiOntologyTerm.getAccession());
```
alf239, line 432: NPE?

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
alf239, line 435: No it doesn't, put a breakpoint and check it out.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
alf239, line 188: My debugger doesn't think so.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
```diff
@@ -0,0 +1,526 @@
...
+
+        Ontology ontology;
+        try {
+            ontology = atlasDAO.getOntologyByName(apiOntology.getName());
+        } catch (DAOException e) { // ontology not found - create a new one
+            ontology = getOrCreateOntology(apiOntology);
+        }
+
+        if (ontology == null) { // ontology already exists - update it
+            ontology.setDescription(apiOntology.getDescription());
```
alf239, line 397: NPE?

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
alf239, line 400: It only creates the ontology, never updates it.

rpetry, line 400: Ah, right, added test

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry, line 435: Tests now cover creation and updating of ontology terms  - I've put a breakpoint on both sections and checked that they are tested.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry, line 188: Added tests to cover this branch

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry, line 432: Thanks - eliminated now.

atlas-web/src/main/java/uk/ac/ebi/gxa/service/CurationService.java
rpetry, line 397: Thanks - eliminated now.

atlas-dao/src/main/java/uk/ac/ebi/gxa/dao/hibernate/DAOException.java
```diff
@@ -0,0 +1,13 @@
+package uk.ac.ebi.gxa.dao.hibernate;
+
+/**
+ * Thrown when DAO's getByName() method returns no results
```
alf239: The class name is _way_ too generic to restrict its usage to _one_ method.

alf239:
Minimised the added code; dropped repeating functionality in the old code where necessary.

@olgamelnichuk, @rpetry — could you please make a final review before the merge?

PR #57, opened 2011-07-28
AtlasProperties is not bound to MyDAS any more

alf239:
It was a weird situation when one of the core classes was bound to the internals of a third-party code. As a side  effect, each test run would create five huge log files impacting text search.

Extracted MyDAS-specific code into MyDAS classes and replaced explicit MyDAS call with `VetoableChangeListener` notification.

### DIFF
--- a/atlas-web/src/main/java/ae3/service/MydasGxaServlet.java
+++ b/atlas-web/src/main/java/ae3/service/MydasGxaServlet.java
@@ -1,11 +1,26 @@
 package ae3.service;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.util.ReflectionUtils;
 import org.springframework.web.context.WebApplicationContext;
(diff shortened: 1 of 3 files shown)


atlas-web/src/main/java/ae3/service/MydasGxaServlet.java
```diff
@@ -13,7 +28,15 @@ import javax.servlet.ServletContext;
  *
  * @see MydasGxaServletContext
  */
-public class MydasGxaServlet extends MydasServlet {
+public class MydasGxaServlet extends MydasServlet implements VetoableChangeListener {
```
alf239, line 35: It would be nice to re-init the `MydasServlet` using the standard `destroy()` and `init()` methods, but unfortunately the settings are stored in a static field which gets destroyed but _not_ cleared, and then not updated during subsequent `init()`. Hence, the reflection tricks are still necessary, which makes the re-init cycle useless.

PR #60, opened 2011-08-02
JS/CSS bundle support

olgamelnichuk:
Please review:
1. added wro4j lib as bundle aggregator
2. created wro4j extension: a jsp tag to put bundles to the jsp
3. Expires and ETag header filters added
4. updated impromptu plugin
5. atlas css is cleaned up a bit


atlas-web/src/main/java/uk/ac/ebi/gxa/web/filter/NoEtagHeaderFilter.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public void doFilter(ServletRequest request, ServletResponse response,
+                         FilterChain chain) throws IOException, ServletException {
+
+        chain.doFilter(request, new HttpServletResponseWrapper((HttpServletResponse) response) {
+            public void setHeader(String name, String value) {
+                if (!"etag".equalsIgnoreCase(name)) {
+                    super.setHeader(name, value);
+                } else {
+                    log.debug("Ignoring Etag header: " + name + " " + value);
```
alf239: Use `{}` please, it's faster in prod mode

olgamelnichuk: Fixed :)

alf239:
It's not forgotten — just postponed until the release is out.

alf239:
This pull request cannot be automatically merged.

atlas-web/src/main/java/uk/ac/ebi/gxa/web/filter/NoEtagHeaderFilter.java
```diff
@@ -0,0 +1,62 @@
...
+ * Ignoring ETag header in tomcat (http://developer.yahoo.com/blogs/ydn/posts/2007/07/high_performanc_11/).
+ *
+ * @author Olga Melnichuk
+ */
+public class NoEtagHeaderFilter implements Filter {
+
+    private static final Logger log = LoggerFactory.getLogger(NoEtagHeaderFilter.class);
+
+    public void init(FilterConfig filterConfig) throws ServletException {
+    }
```
alf239, line 44: log please

atlas-wro4j/src/main/java/uk/ac/ebi/gxa/web/wro4j/tag/AggregatedResourcePath.java
```diff
@@ -0,0 +1,80 @@
...
+    private final AggregatedResourceNamePattern namePattern;
+    private final String aggregationPath;
+
+    public AggregatedResourcePath(@Nullable String aggregationPath, AggregatedResourceNamePattern namePattern) {
+        this.aggregationPath = aggregationPath == null ? "" : aggregationPath;
+        this.namePattern = namePattern;
+    }
+
+    public ResourceFinder relativeTo(PageContext pageContext) {
+        String contextRealPath = pageContext.getServletContext().getRealPath("/");
```
alf239: Unfortunately, `getRealPath` can well return `null` even for proper resources:

> The real path returned will be in a form appropriate to the computer and operating system on which the servlet container is running, including the proper path separators. This method returns `null` if the servlet container cannot translate the virtual path to a real path for any reason (such as when the content is being made available from a `.war` archive).

That makes this approach not suitable for real life.

alf239: [`ServletContext.getResourcePaths(java.lang.String)`](http://download.oracle.com/javaee/1.3/api/javax/servlet/ServletContext.html) should do what we need here; still, that would mean the file filtering approach is to be adjusted

alf239:
We still have

``` html
        <link rel="stylesheet" href="http://www.ebi.ac.uk/inc/css/contents.css" type="text/css"/>
        <link rel="stylesheet" href="http://www.ebi.ac.uk/inc/css/userstyles.css" type="text/css"/>
        <link rel="stylesheet" href="/gxa/atlas-ebi.css" type="text/css" type="text/css"/>
        <script src="http://www.ebi.ac.uk/inc/js/contents.js" type="text/javascript"></script>
```

non-aggregated and uncompressed

alf239:
@rpetry, @ostolop — could you please review? I've incorporated my notes and covered the code with tests to 87% coverage

I would not spend time on the CSS part, as the potential bugs are trivial; the question is the tag itself: I've changed it a bit, so cannot just merge it without further review :)

atlas-wro4j/src/test/resources/config/wro-with-cycle-refs.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<groups xmlns="http://www.isdc.ro/wro"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.isdc.ro/wro wro.xsd">
```
alf239: Cannot see usages of this file. Is it necessary to keep it around?

It looks as if it was a test for Wro4j itself, proving that our assumptions stand. Could we please create that test if it's still important?

atlas-wro4j/src/main/java/uk/ac/ebi/gxa/web/wro4j/tag/AggregatedResourcePath.java
alf239: fixed

PR #69, opened 2011-08-08
Ticket #3051 Gene page: number of expression profiles in all studies is aways zero (pre-release 2.0.9)

olgamelnichuk:
Temporary fix: EfvAttribute for empty ef/efv pair should be null when getting the number of experiments.

ready to review


alf239:
For further quick fixes, please update the tickets with links to quick fix and describe further steps.

PR #72, opened 2011-08-09
Trac3052

ostolop:
@alf239 -- could you review this please?

### DIFF
--- a/atlas-analytics/src/main/java/uk/ac/ebi/gxa/analytics/generator/service/ExperimentAnalyticsGeneratorService.java
+++ b/atlas-analytics/src/main/java/uk/ac/ebi/gxa/analytics/generator/service/ExperimentAnalyticsGeneratorService.java
@@ -61,7 +61,8 @@ public class ExperimentAnalyticsGeneratorService {
     private final Logger log = LoggerFactory.getLogger(this.getClass());
     private ExecutorService executor;
 
-    public ExperimentAnalyticsGeneratorService(AtlasDAO atlasDAO, AtlasDataDAO atlasDataDAO, AtlasComputeService atlasComputeService, ExecutorService executor) {
+    public ExperimentAnalyticsGeneratorService(AtlasDAO atlasDAO, AtlasDataDAO atlasDataDAO,
+                                               AtlasComputeService atlasComputeService, ExecutorService executor) {
         this.atlasDAO = atlasDAO;
         this.atlasDataDAO = atlasDataDAO;
         this.atlasComputeService = atlasComputeService;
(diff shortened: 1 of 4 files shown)


atlas-analytics/src/main/java/uk/ac/ebi/gxa/analytics/generator/service/ExperimentAnalyticsGeneratorService.java
```diff
@@ -161,8 +161,8 @@ public class ExperimentAnalyticsGeneratorService {
         for (NetCDFDescriptor netCDF : netCDFs) {
             count++;
 
-            if (!factorsCharacteristicsAvailable(netCDF)) {
-                listener.buildWarning("No analytics were computed for " + netCDF + " as it contained no factors or characteristics!");
+            if (!factorsAvailable(netCDF)) {
+                listener.buildWarning("No analytics were computed for " + experimentAccession + " as it contained no factors!");
```
alf239: Isn't it misleading in case an Experiment has more than one NetCDF?

ostolop: It's possible... My change only concerned making the scope of the check smaller, "factor" vs "factor & characteristic". 

In the netcdf-to-merge branch the log message is improved -- experimentAccession & arrayDesign are both output. As it stands here, the NetCDFDescriptor has no ad accession. Best to leave as is it.

alf239: Could we please stick with `listener.buildWarning("No analytics were computed for " + netCDF + " as it contained no factors!");` then? I don't want to make log messages even more misleading unless we have a _really_ good reason to do so.

atlas-analytics/src/test/java/uk/ac/ebi/gxa/analytics/generator/TestExperimentAnalyticsGeneratorService.java
```diff
@@ -22,25 +22,52 @@
...
+                .andThrow(new AssertionFailedError("Unexpected call to computeTask"))
+                .anyTimes();
+        replay(atlasComputeService);
 
-        // create a reader to read in code
-        BufferedReader reader = new BufferedReader(new InputStreamReader(in));
+        final ExperimentAnalyticsGeneratorService experimentAnalyticsGeneratorService =
+                new ExperimentAnalyticsGeneratorService(
+                atlasDAO, atlasDataDAO, atlasComputeService,
+                EasyMock.<ExecutorService>anyObject());
```
alf239: It should be `createMock(ExecutorService.class)` actually. It only works now because ExecutorService is not used.

PR #74, opened 2011-08-11
Associations done right

alf239:
Association table defining many-to-many relation should not have its own ID.


alf239:
Not relevant for 2.0.9

PR #76, opened 2011-08-11
Trac3057

ostolop:
@alf239 please review

### DIFF
--- a/atlas-data-storage/src/main/java/uk/ac/ebi/gxa/data/NetCDFProxy.java
+++ b/atlas-data-storage/src/main/java/uk/ac/ebi/gxa/data/NetCDFProxy.java
@@ -362,7 +362,7 @@ public final class NetCDFProxy implements Closeable {
         return result;
     }
 
-    public List<KeyValuePair> getUniqueValues() throws IOException {
+    public List<KeyValuePair> getUniqueValues() throws IOException, AtlasDataException {
         Variable uVALVar;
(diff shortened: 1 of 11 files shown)


atlas-data-storage/src/main/java/uk/ac/ebi/gxa/data/NetCDFProxy.java
```diff
@@ -378,17 +378,20 @@ public final class NetCDFProxy implements Closeable {
...
+        for (Object text : (Object[]) uVal.make1DStringArray().get1DJavaArray(String.class)) {
+	        final String[] data = ((String) text).split(NCDF_PROP_VAL_SEP_REGEX, -1);
             if (data.length != 2) {
-                throw new RuntimeException("Invalid uVAL element: " + text);
+                throw new AtlasDataException("Invalid uVAL element: " + text);
+            }
+
+            if (!"".equals(data[1])) {
+                list.add(new KeyValuePair(data[0], data[1]));
             }
```
alf239, line 389: Am I right that we silently ignore all the `uVAL`s where `"".equals(data[1])`? Is it as intended?

ostolop, line 389: Correct. It means that the uVAL was like "foo||" -- i.e. had an empty second value. These we ignore.

PR #78, opened 2011-08-16
Getting rid of hilo strategy in Hibernate's annotations

alf239:
- setting `allocationSize = 1` in order to disable Hibernate's hilo approach (which is nice, but not ready for our DB yet)
- adding script to update `ATLASMGR` (it was apparently out-of-date in curators' instance)
- as the bug with `hilo` approach could get sequences out of sync with the tables, it won't hurt running `RebuildSequences()` just to be sure (it's idempotent, anyway)
- to facilitate the merge, made `arrayDesignID` `Long` rather than `long`

### DIFF
--- a/atlas-dao/src/test/java/uk/ac/ebi/gxa/dao/TestAtlasDAO.java
+++ b/atlas-dao/src/test/java/uk/ac/ebi/gxa/dao/TestAtlasDAO.java
@@ -123,8 +123,8 @@ public class TestAtlasDAO extends AtlasDAOTestCase {
         String accession =
                 getDataSet().getTable("A2_ARRAYDESIGN").getValue(0, "accession")
                         .toString();
-        long id =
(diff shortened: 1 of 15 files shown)


alf239:
The documentation on this is vague to say the least.

``` java
    /**
     * (Optional) The amount to increment by when allocating
     * sequence numbers from the sequence.
     */
    int allocationSize() default 50;
```

In case of Hibernate, unless you're using `hibernate.id.new_generator_mappings`, the sequence will be used as _hi_ part of hilo approach: Hibernate will get the value from sequence (incremented by one), multiply it by `allocationSize`, and make its own ids from this value onwards to this value plus 49. That's all good, but that's not the way we use sequences in Oracle. Hence, until we've dropped the triggers, we must keep the behaviour consistent between Hibernate and triggers, and so we do.

Note though that it can and will be [a bottleneck](https://forum.hibernate.org/viewtopic.php?p=2422062).

PR #86, opened 2011-08-17
Ticket#3063 Switching AD on experiment page is broken

olgamelnichuk:
just a typo


alf239:
[RFC 1738](http://www.ietf.org/rfc/rfc1738.txt) seems to be happy with that. Not so sure about browsers...

alf239:
The `search` property returns the query portion of a URL, _including the question mark (`?`)._

PR #87, opened 2011-08-18
Trac3066

ostolop:
Fixes #3066 -- filtering on EFVs is done in the builder, not in NetCDFProxy. This should merge into server-curators. There's a separate pull request to merge into develop.

### DIFF
--- a/indexbuilder/src/main/java/uk/ac/ebi/gxa/index/builder/service/GeneAtlasBitIndexBuilderService.java
+++ b/indexbuilder/src/main/java/uk/ac/ebi/gxa/index/builder/service/GeneAtlasBitIndexBuilderService.java
@@ -138,9 +138,11 @@ public class GeneAtlasBitIndexBuilderService extends IndexBuilderService {
 
                 final ExperimentInfo experiment = experimentPool.intern(new ExperimentInfo(exp.getAccession(), exp.getId()));
 
-                // TODO when we switch on inclusion of sc-scv stats in bit index, the call below
-                // TODO should change to ncdf.getUniqueValues()
-                List<String> uVals = ncdf.getUniqueFactorValues();
+                final List<String> uVals = ncdf.getUniqueValues();
+
+                // TODO to switch on inclusion of sc-scv stats in bit index, remove getFactors & !contains filter below
+                final Set<String> factorNames = new HashSet<String>(Arrays.asList(ncdf.getFactors()));
+
                 int car = 0; // count of all Statistics records added for this ncdf
 
                 if (uVals.size() == 0) {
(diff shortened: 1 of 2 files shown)


ostolop:
@alf239

indexbuilder/src/main/java/uk/ac/ebi/gxa/index/builder/service/GeneAtlasBitIndexBuilderService.java
```diff
@@ -156,9 +158,17 @@ public class GeneAtlasBitIndexBuilderService extends IndexBuilderService {
 
                 final Map<EfvAttribute, MinPMaxT> efToPTUpDown = new HashMap<EfvAttribute, MinPMaxT>();
                 for (int j = 0; j < uVals.size(); j++) {
-                    String[] arr = uVals.get(j).split(NetCDFProxy.NCDF_PROP_VAL_SEP_REGEX);
-                    String ef = internedCopy(stringPool, arr[0]);
-                    String efv = arr.length == 1 ? "" : internedCopy(stringPool, arr[1]);
+                    final String[] arr = uVals.get(j).split(NetCDFProxy.NCDF_PROP_VAL_SEP_REGEX, -1);
+
+                    if (arr.length != 2 || "".equals(arr[1]) || "(empty)".equals(arr[1]))
+                        continue;
```
alf239, line 164: `NetCDFProxy.getUniqueFactorValues()` seems to be closely related, but we do not adjust its filtering, why?

``` java

    public List<String> getUniqueFactorValues() throws IOException {
        List<String> uniqueEFVs = new ArrayList<String>();
        List<String> factors = Arrays.asList(getFactors());

        for (String propVal : getUniqueValues()) {
            String[] nameValue = propVal.split(NCDF_PROP_VAL_SEP_REGEX, -1);
            if (factors.contains(nameValue[0]) && !"".equals(nameValue[1])) {
                // Since getUniqueValues() returns both ef-efvs/sc-scvs, filter out scs that aren't also efs
                uniqueEFVs.add(propVal);
            }
        }
        return uniqueEFVs;
    }
```

alf239:
To avoid further confusion: the unique values are part of NetCDF structure, which means that indices of `uVals` are important. Hence, in order for the lines 

``` java
                        float t = tstat.get(i, j);
                        float p = pvals.get(i, j);
```

to work correctly we must have `j` referring to non-filtered `uVals`. Scary, but best we can do given the timeframe. It could be avoided using tuples _(coordinate, value)_ instead of value itself, but whether or not it's more clear is a subject for a separate discussion.

indexbuilder/src/main/java/uk/ac/ebi/gxa/index/builder/service/GeneAtlasBitIndexBuilderService.java
alf239, line 164: It's a separate [ticket 2958](http://bar.ebi.ac.uk:8080/trac/ticket/2958), not too important at the moment.

PR #92, opened 2011-08-19
Ticket #3071 Weird rounding in JSON/XML API output

olgamelnichuk:
Floating point number representation is not exact for some numbers, e.g. 0.1, so using BigDecimal now.


olgamelnichuk:
http://bar.ebi.ac.uk:8080/trac/ticket/3071

PR #96, opened 2011-08-19
Ticket #2988 EB-eye Search Link needs to be replaced with smth more current

olgamelnichuk:
Search EB-eye link is set to  http://www.ebi.ac.uk/ebisearch/search.ebi?db=allebi&query=${atlasGene.geneIdentifier}&requestFrom=ebi_index&submit=+FIND+.


olgamelnichuk:
http://bar.ebi.ac.uk:8080/trac/ticket/2988

PR #100, opened 2011-08-19
Fixed indexing of design elements for genes

nsklyar:


ostolop:
Won't merge -- this is not the right branch.

PR #102, opened 2011-08-19
Ticket #2911 Gene search broken on experiment page

olgamelnichuk:
Gene search returned default list of genes when junk text is given and array design is specified.


alf239:
@geometer: please note the change: it might be painful to merge

PR #107, opened 2011-08-25
Ticket #2911 Enable full-text gene search on experiment page

olgamelnichuk:
bugfix: when the gene query is not empty but no genes found an empty list should be shown
http://bar.ebi.ac.uk:8080/trac/ticket/2911

@alf239 please review


alf239:
Ok by me; still, I cannot say I understand results for e.g. `INS INS5`, especially given that these are not consistent with `INS5` and `INS` issued separately.

alf239:
Please review my corrections/fixes as well.

PR #111, opened 2011-09-01
Ticket #3091 Efvs with containing single quotes should be escaped when server-side javascript is being generated

olgamelnichuk:
Applied javascript escape utility function.

@rpetry - please, review


olgamelnichuk:
http://bar.ebi.ac.uk:8080/trac/ticket/3091

PR #113, opened 2011-09-01
Ticket #2929 if "Autodetect" is not selected in the Tasks drop-down on Admin, checkboxes are ignored

olgamelnichuk:
@rpetry - please, review


olgamelnichuk:
http://bar.ebi.ac.uk:8080/trac/ticket/2929

PR #114, opened 2011-09-01
Ticket #3069 Invalid solr query "id:()" when search by existing gene but there is no experiments for it

olgamelnichuk:
@rpetry - please, review


olgamelnichuk:
http://bar.ebi.ac.uk:8080/trac/ticket/3096

PR #121, opened 2011-09-08
Transactions

alf239:
As per http://forum.springsource.org/showthread.php?57763-Transactional-on-private-methods 

  Only external method calls can be intercepted (as explained in [the reference guide](http://static.springsource.org/spring/docs/3.0.5.RELEASE/reference/transaction.html#transaction-declarative-annotations)) so marking private methods with `@Transactional` doesn't do anything.

See the  [10.5.6 Using `@Transactional`](http://static.springsource.org/spring/docs/3.0.5.RELEASE/reference/transaction.html#transaction-declarative-annotations):

You can place the `@Transactional` annotation before an interface definition, a method on an interface, a class definition, or a _public_ method on a class.


alf239:
oops, wrong one.

PR #125, opened 2011-09-13
Trac3106

rpetry:
@olgamielnichuk - please review


alf239:
@olgamielnichuk doesn't link to a user; @olgamelnichuk works better

olgamelnichuk:
Heatmap cell gives us the ef/efv pair already, why should we search for maximum pValue?

rpetry:
I take it you meant min pValue. But this problem is affecting efo cells, rather than ef-efv. Whereas the plot legend displays ef-efvs, and sometimes an efo maps to more than one ef-efv in a given experiment -  as I described in the ticket.

PR #130, opened 2011-09-16
Trac 2835 Automatic DB updates

alf239:
Trac ticket #[2835 Implement schema update when new software is installed](http://bar.ebi.ac.uk:8080/trac/ticket/2835). 
- Run during startup
- Based on CarbonFive db-migrator
- With a brand new Oracle SQL splitter (shame on me—could not find anything even closely reasonable, so created yet another beast)
- Currently SQL only, while in theory can support Groovy migrations as well, see the `groovy` branch
- The very first migration checks the `A2_SCHEMACHANGES` is in place—and drops it. Thus, if it was the first start, we make sure we're running against 2.0.9-compatible DB. If it's not the first start, the corresponding change was already run and logged, hence it won't run again
- Release package contains change history now

TODO items:
- The whole release packaging should be reviewed and tested twice


alf239:
@rpetry, @olgamelnichuk, @ostolop — fancy a review?

alf239:
Fixes http://bar.ebi.ac.uk:8080/trac/ticket/2835

alf239:
Sample log:

```
2011-09-16 15:41:20,413 INFO  [main] migration.DataSourceMigrationManager (DataSourceMigrationManager.java:191) - Successfully enabled migrations.
2011-09-16 15:41:20,480 INFO  [main] migration.DataSourceMigrationManager (DataSourceMigrationManager.java:76) - Migrating database... applying 3 migrations.
2011-09-16 15:41:20,482 INFO  [main] migration.DataSourceMigrationManager$1 (DataSourceMigrationManager.java:94) - Running migration 201101010001-initial-state-check.sql.
2011-09-16 15:41:20,487 INFO  [main] db.SqlStatementExecutor (SqlStatementExecutor.java:25) - select * from a2_schemachanges
2011-09-16 15:41:20,552 INFO  [main] db.SqlStatementExecutor (SqlStatementExecutor.java:25) - drop table a2_schemachanges
2011-09-16 15:41:21,019 INFO  [main] migration.DataSourceMigrationManager$1 (DataSourceMigrationManager.java:94) - Running migration 201108160001-add-ontology-seq.sql.
2011-09-16 15:41:21,021 INFO  [main] db.SqlStatementExecutor (SqlStatementExecutor.java:25) - CREATE SEQUENCE  "A2_ONTOLOGY_SEQ"
2011-09-16 15:41:21,040 INFO  [main] db.SqlStatementExecutor (SqlStatementExecutor.java:25) - CREATE OR REPLACE TRIGGER A2_ONTOLOGY_INSERT before insert on A2_Ontology for each row begin if (:new.OntologyID is null) then select A2_Ontology_seq.nextval into :new.OntologyID from dual; end if; end;
2011-09-16 15:41:21,221 INFO  [main] db.SqlStatementExecutor (SqlStatementExecutor.java:25) - ALTER TRIGGER A2_ONTOLOGY_INSERT ENABLE
2011-09-16 15:41:21,239 INFO  [main] db.SqlStatementExecutor (SqlStatementExecutor.java:25) - call atlasmgr.RebuildSequences()
2011-09-16 15:41:35,831 INFO  [main] migration.DataSourceMigrationManager$1 (DataSourceMigrationManager.java:94) - Running migration 201108160002-drop-assaysample-seq.sql.
2011-09-16 15:41:35,833 INFO  [main] db.SqlStatementExecutor (SqlStatementExecutor.java:25) - DROP TRIGGER "A2_ASSAYSAMPLE_INSERT"
2011-09-16 15:41:35,940 INFO  [main] db.SqlStatementExecutor (SqlStatementExecutor.java:25) - DROP SEQUENCE "A2_ASSAYSAMPLE_SEQ"
2011-09-16 15:41:35,962 INFO  [main] db.SqlStatementExecutor (SqlStatementExecutor.java:25) - ALTER TABLE "A2_ASSAYSAMPLE" DROP CONSTRAINT PK_ASSAYSAMPLE
2011-09-16 15:41:36,092 INFO  [main] db.SqlStatementExecutor (SqlStatementExecutor.java:25) - ALTER TABLE "A2_ASSAYSAMPLE" DROP CONSTRAINT UQ_ASSAYSAMPLE
2011-09-16 15:41:36,204 INFO  [main] db.SqlStatementExecutor (SqlStatementExecutor.java:25) - ALTER TABLE "A2_ASSAYSAMPLE" DROP COLUMN ASSAYSAMPLEID
2011-09-16 15:41:45,637 INFO  [main] db.SqlStatementExecutor (SqlStatementExecutor.java:25) - ALTER TABLE "A2_ASSAYSAMPLE" ADD CONSTRAINT "PK_ASSAYSAMPLE" PRIMARY KEY (ASSAYID,SAMPLEID) /*PK_TABLESPACE*/ ENABLE
2011-09-16 15:41:46,607 INFO  [main] migration.DataSourceMigrationManager (DataSourceMigrationManager.java:130) - Migrated database in 0:00:26.122.
```

atlas-updates/src/main/java/uk/ac/ebi/gxa/db/OracleScriptSplitter.java
```diff
@@ -0,0 +1,102 @@
...
+                   Terminate PL/SQL subprograms by entering a period (.) by itself on a new line.
+                   You can also terminate and execute a PL/SQL subprogram by entering a slash (/)
+                   by itself on a new line.
+                    */
+                    execute(executor, sqlBuffer);
+                    plsqlMode = false;
+                    sqlBuffer = null;
+                } else if (!plsqlMode && sqlBuffer.toString().toLowerCase().matches("(" +
+                        "begin|" +
+                        "declare|" +
```
alf239: SQL_Plus switches to PL/SQL mode on `begin`/`declare`, not necessary as the very first statement. `(._\s)?(begin|declare)\s.*` would be more like the spec says. OTOH, I need to find a proper counterexample first.

alf239:
Two problems with production:
1. The change will try to update production DB which is strange, as we used to hare read-only prod
2. The change was not tested in cluster

atlas-updates/src/main/resources/db/migrations/201101010002-atlasmgr.sql
```diff
@@ -0,0 +1,156 @@
```
rpetry: Might it be worth reflecting the changes in PKG_ATLASMGR.sql?

alf239: I'd prefer to add `/*`-style comments support :)) Seriously, the only change is lack of comments; I thought it'd make `PKG_ATLASMGR.sql` worse than it was before.

alf239: On a second thought, I think I can make it pretty enough and still the same. Will do, thank you.

atlas-updates/src/main/resources/db/migrations/201108160001-add-ontology-seq.sql
```diff
@@ -4,15 +4,12 @@ CREATE OR REPLACE TRIGGER A2_ONTOLOGY_INSERT
```
rpetry, line 3: Now that the migrations would be happening at start up, would it be possible to make adding new objects more resilient, e.g. CREATE SEQUENCE  "A2_ONTOLOGY_SEQ"? Perhaps by running select \* from USER_SEQUENCES where sequence_name = 'A2_ONTOLOGY_SEQ' and dropping if necessary - I happened to have the schema I was migrating and 201108160001-add-ontology-seq.sql step failed.

atlas-updates/src/main/resources/db/migrations/201101010002-atlasmgr.sql
```diff
@@ -0,0 +1,156 @@
+CREATE OR REPLACE PACKAGE ATLASMGR IS
```
rpetry: It's interesting, but once the 201101010002-atlasmgr.sql step succeeded I looked at ATLASMGR package in my schema and saw that the whole package body was in one line,  hence starting from the first comment '--orphane ev in release..' the remainder of the body was essentially commented out also.

atlas-updates/src/test/resources/uk/ac/ebi/gxa/db/TestOracleScriptSplitter-sample.sql
```diff
@@ -0,0 +1,15 @@
+DROP TRIGGER "A2_ASSAYSAMPLE_INSERT";
+
+DROP SEQUENCE "A2_ASSAYSAMPLE_SEQ";
```
rpetry, line 3: Again, could the above two statements first check if the object exists before dropping it? Perhaps we could have a common stored proc that drops a DB object only if that object exists, and pass to that proc a &lt;name&gt; that identifies which user_ &lt;name&gt; table to check the object existence in?

atlas-updates/src/main/resources/db/migrations/201101010002-atlasmgr.sql
alf239: That's due to our inability to support end-of-line comments as `--orphane ev in release 10.3` below. Fixed and renamed the file so that it would get updated on the next attempt.

Supporting `/*`-style and end-of-line comments requires a proper lexer (in order to avoid commenting out parts of strings for example, like `where x like '%--%'` — I'm not sure we're able to support lexer in scope of Atlas.

Yet another problem is, there's no ready-to-use grammatic for SQL, PL/SQL, and sqlplus commands. The best ones I could find are limited to a subset of PL/SQL.

atlas-updates/src/main/resources/db/migrations/201108160001-add-ontology-seq.sql
alf239, line 3: I'm not sure we want this. If we're trying to remove a sequence which is not in DB, it means our assumptions about the database are wrong; I see no harm in failing in such a situation. 

As an alternative, we could require all the changes to be idempotent, but I see no way to ensure that.

atlas-updates/src/test/resources/uk/ac/ebi/gxa/db/TestOracleScriptSplitter-sample.sql
alf239, line 3: Irrelevant here — these are just examples of code to check the splitter works reasonably well (it's obviously bad, as ad hoc parser cannot be good—but it's the best we can get relatively fast)

atlas-updates/src/main/resources/db/migrations/201108160001-add-ontology-seq.sql
rpetry, line 3: We agreed that re-runnability of statements would be nice, but as it requires more development and obfuscates the fact that our assumptions about the DB we're updating are wrong - we've decided to leave it to one side for now - at least until its lack proves painful. Then we can review the reasons why we'd need it and make a decision.

atlas-updates/src/main/resources/db/migrations/201101010002-atlasmgr.sql
alf239: Returned multi-line statements; that should actually fix end-of-line comments, too: Oracle is capable of recognising comments in package code, so as long as we split statements properly, we're fine.

atlas-updates/src/main/java/uk/ac/ebi/gxa/db/OracleScriptSplitter.java
alf239: Addressed. We only support `begin` and `declare` as a separate line so far (even comments will not be allowed there), but it should be enough for our purposes.

PR #134, opened 2011-09-20
Trac3112

rpetry:
Ticket #3112 - Direct single gene queries to heatmap instead of the gene page


alf239:
Fixes http://bar.ebi.ac.uk:8080/trac/ticket/3112

alf239:
Looks fine, also fixes an infinite-redirect bug.

PR #135, opened 2011-09-20
Following up trac 3112

alf239:
- fixed and clear list of parameters
- no parameter overrides


alf239:
@rpetry, please take a look

PR #139, opened 2011-09-23
Rpetry

rpetry:

### DIFF
--- a/atlas-updates/src/main/resources/db/migrations/201108160002-drop-assaysample-seq.sql
+++ b/atlas-updates/src/main/resources/db/migrations/201108160002-drop-assaysample-seq.sql
@@ -13,3 +13,7 @@ ALTER TABLE "A2_ASSAYSAMPLE"
     PRIMARY KEY (ASSAYID,SAMPLEID)
     /*PK_TABLESPACE*/
     ENABLE;  
+
+CREATE OR REPLACE VIEW "VWSAMPLEASSAY" ("ASSAYID", "SAMPLEID", "ASSAYACCESSION") AS select asa.ASSAYID, asa.SAMPLEID, a.Accession AssayAccession
+from a2_assaysample asa
+join a2_assay a on a.AssayID = asa.AssayID;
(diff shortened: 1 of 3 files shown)


atlas-updates/src/main/resources/db/migrations/201108160002-drop-assaysample-seq.sql
```diff
@@ -13,3 +13,7 @@ ALTER TABLE "A2_ASSAYSAMPLE"
     PRIMARY KEY (ASSAYID,SAMPLEID)
     /*PK_TABLESPACE*/
     ENABLE;  
+
+CREATE OR REPLACE VIEW "VWSAMPLEASSAY" ("ASSAYID", "SAMPLEID", "ASSAYACCESSION") AS select asa.ASSAYID, asa.SAMPLEID, a.Accession AssayAccession
+from a2_assaysample asa
+join a2_assay a on a.AssayID = asa.AssayID;
```
alf239: Editing updates is a tricky approach: the developers who already run this script will not get or even notice the update. For idempotent updates, renaming the file so that ID would change works fine, but in our case `DROP` will fail.

So it would be better to have it as a separate change. In this particular case, though, it's not too important, as `VWSAMPLEASSAY` is not supposed to be used by developers.

atlas-updates/src/main/resources/db/migrations/201109231350-tr-cur-ontologymapping.sql
```diff
@@ -0,0 +1,200 @@
+/*******************************************************************************
+select * from CUR_OntologyMapping
+*******************************************************************************/
```
alf239, line 3: Please check whether this part works — I did add some kind of support for `/*`-style comments, but have never actually tested it.

atlas-updates/src/main/resources/db/migrations/201108160002-drop-assaysample-seq.sql
rpetry: Agreed - moved out to a different file.

atlas-updates/src/main/resources/db/migrations/201109231350-tr-cur-ontologymapping.sql
rpetry, line 3: I have tested - it does - see A2_schemaversion in my schema.

PR #148, opened 2011-10-04
Trac 2745: a2_ontologymapping.ontologytermname is always null

alf239:
Trac 2745: `a2_ontologymapping.ontologytermname` is always `null`

Dropped the column

http://bar.ebi.ac.uk:8080/trac/ticket/2745


alf239:
Right, fixed. Thanks for spotting!

rpetry:
Tested the migration file against a 2.0.10 schema

PR #149, opened 2011-10-04
Trac3106

rpetry:
@olgamelnichuk - could you review please. Apologies for closing [the previous pull request](https://github.com/gxa/gxa/pull/125), #125, by mistake. Please see my last comment in that request:

> I take it you meant min pValue. But this problem is affecting efo cells, rather than ef-efv. Whereas the plot legend displays ef-efvs, and sometimes an efo maps to more than one ef-efv in a given experiment—as I described in [the ticket](http://bar.ebi.ac.uk:8080/trac/ticket/3106).


alf239:
added a link to the #125 and a quote of the comment in question

olgamelnichuk:
I'm still not quite sure what we show in the experiments popup. The bar plot uses given ef/efv just to define the correct array design, but do not affect the plot itself. So if the ef/efv with the best PValue is not at the top of the legend than this change is useless..

PR #151, opened 2011-10-07
Gene Page minor bug fixes 

olgamelnichuk:
1. no more overlapping in EF pagination (Trac 2684)
2. no more plot tooltips outside of plots (Trac 2685)
3. click on ef pagination do not lost yellow efv selection on the plot 
4. al js code moved out from jsp to easy future development


rpetry:
I've tested the test cases listed above and examined the changes in Java. I've looked at the js/jsp changes at a certain level of abstraction only.

PR #152, opened 2011-10-10
Trac 3132: AIOOBE 

alf239:
Added non-filtered `getUVals` as a temporary solution; as far as I can see, the `PVAL` and `TSTAT` structures are still parallel to `uVAL`, so that should be enough for a quick relief.

For the next step, we must either get rid of or encapsulate the parallel structures we use.

### DIFF
--- a/atlas-data-storage/src/main/java/uk/ac/ebi/gxa/data/NetCDFProxy.java
+++ b/atlas-data-storage/src/main/java/uk/ac/ebi/gxa/data/NetCDFProxy.java
@@ -367,6 +367,32 @@ public final class NetCDFProxy implements Closeable {
         return list;
     }
 
+    public List<KeyValuePair> getUVal() throws IOException, AtlasDataException {
+        Variable uVALVar;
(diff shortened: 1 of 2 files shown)


atlas-data-storage/src/main/java/uk/ac/ebi/gxa/data/NetCDFProxy.java
```diff
@@ -367,6 +367,32 @@ public final class NetCDFProxy implements Closeable {
         return list;
     }
 
+    public List<KeyValuePair> getUVal() throws IOException, AtlasDataException {
```
rpetry, line 385: Could we have a TODO comment (preferable a new trac number) in the code - given that this method looks almost identical to the one above?

alf239, line 385: Here you are

alf239:
For further changes, refer to [Trac ticket 3150](http://bar.ebi.ac.uk:8080/trac/ticket/3150)

PR #153, opened 2011-10-11
Trac3154

rpetry:
[Ticket 3154](http://bar.ebi.ac.uk:8080/trac/ticket/3154) — hot fix to prevent UQ_PROPERTY_NAME constraint vlolation exception. A more complete solution to this issue will be implemented in develop as part of [Ticket 2849](http://bar.ebi.ac.uk:8080/trac/ticket/2849)


alf239:
Should be fine as a temporary solution; checking just in case.

PR #158, opened 2011-10-19
Trac3162

rpetry:

### DIFF
--- a/scripts/cron/public_private_ae2_to_atlas.sh
+++ b/scripts/cron/public_private_ae2_to_atlas.sh
@@ -60,7 +60,7 @@ while read line; do
           # E.g. line in $all_ae2_experiments_file: accession:E-MEXP-31 privacy:public releasedate:2004-03-01
           ae2_experiment=`grep "accession:$exp_accession " $all_ae2_experiments_file`
 	      ae2_release_date=`echo $ae2_experiment | awk '{print $3}' | cut -d':' -f2`
-          if [ ! -z ae2_experiment ]; then
+          if [ ! -z $ae2_experiment ]; then
                ae2_public_status=`echo $ae2_experiment | grep -Po 'privacy:public'`
                ae2_private_status=`echo $ae2_experiment | grep -Po 'privacy:private'`
                if [ ! -z $ae2_public_status ]; then
--- a/scripts/cron/public_private_ae2_to_atlas.sh
+++ b/scripts/cron/public_private_ae2_to_atlas.sh
@@ -1,0 +1,101 @@
+#!/bin/bash
+# @author: rpetry
+# @date:   18 Oct 2011
+
+# This script serves to reflect in Atlas private/public experiment status switch on the AE2 side.
+
+if [ $# -eq 0 ]; then
+        echo "Usage: $0 ADMIN_USERNAME ADMIN_PASSWORD ATLAS_URL ATLAS_PORT ATLAS_ROOT ERROR_NOTIFICATION_EMAILADDRESS"
+        echo "e.g. $0 admin <pwd> lime 14032 gxa-load atlas-developers@ebi.ac.uk"
+        exit;
+fi
+
+process_data="$3.$4.`eval date +%Y%m%d`"
+process_file="/tmp/privatepublic_ae2_to_atlas.$process_data"
+authentication_cookie=$process_file.$$
+all_atlas_experiments_file=$process_file.$$.all_exps
+all_atlas_experiments_file=$process_file.$$.all_atlas_exps
+all_ae2_experiments_file=$process_file.$$.all_ae2_exps
+
+# Remove any previous $process_file
+rm -rf $process_file
+
+# login to Atlas admin
+curl -X GET -c $authentication_cookie -H "Accept: application/json" "http://$3:$4/$5/admin?op=login&userName=$1&password=$2&indent" >> /dev/null
+
+# cut commands extract e.g. 0 from {"numTotal":0,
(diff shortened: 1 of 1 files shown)


scripts/cron/public_private_ae2_to_atlas.sh
```diff
@@ -0,0 +1,105 @@
...
+# Retrieve all Atlas experiments into $all_atlas_experiments_file (json)
+curl -X GET -b $authentication_cookie -H "Accept: application/json" "http://$3:$4/$5/admin?&p=0&n=$num_all_atlas_exps&search=&fromDate=&toDate=&pendingOnly=ALL&op=searchexp&indent" > $all_atlas_experiments_file
+if [ ! -f $all_atlas_experiments_file ]; then
+   err_msg="Updating private/public status of experiments on $3:$4/$5 was unsuccessful due failure to retrieve all Atlas experiments"
+   echo $err_msg >> $process_file.log
+   mailx -s "$err_msg" $6 < $process_file.log
+fi
+
+# Retrieve all AE2 experiments into $all_ae2_experiments_file (ssv)
+curl -X GET "http://www.ebi.ac.uk/arrayexpress/admin/privacy" > $all_ae2_experiments_file
```
olgamelnichuk, line 39: Should the arrayexpress privacy url http://www.ebi.ac.uk/arrayexpress/admin/privacy be working already? curl command returns "The resource you are requesting is not found..." to me.

rpetry, line 39: Sorry Olga, I should have commented with the pull request - the functionality on the AE2 side is not in www yet - but can be tested with wwwdev (wwwdev points at prod AE2 DB also).

scripts/cron/public_private_ae2_to_atlas.sh
```diff
@@ -0,0 +1,105 @@
...
+   val=`echo $line | sed 's|[\", ]||g' | grep -P 'private' | cut -d':' -f2`
+   if [ ! -z $val ]; then
+       atlas_private_flag=$val
+
+       if [ ! -z $exp_accession ]; then
+          # Now get the experiment's private/public status in AE2
+          # E.g. line in $all_ae2_experiments_file: accession:E-MEXP-31 privacy:public releasedate:2004-03-01
+          ae2_experiment=`grep "accession:$exp_accession " $all_ae2_experiments_file`
+	      ae2_release_date=`echo $ae2_experiment | awk '{print $3}' | cut -d':' -f2`
+          if [ ! -z ae2_experiment ]; then
```
olgamelnichuk: Looks like a typo; ae_experiment variable without "$". Probably it should be like this:

```
if [ ! -z $ae2_experiment ]; then 
   ...
fi
```

rpetry: Well spotted - thanks Olga - fixed.

scripts/cron/public_private_ae2_to_atlas.sh
```diff
@@ -0,0 +1,105 @@
...
+                   if [ $atlas_private_flag == "false" ]; then
+                      # Experiment private in AE2 and public in Atlas - make it private in Atlas
+                      echo "$exp_accession - AE2: private; Atlas: public - status change in Atlas: public->private" >> $process_file.log
+                      curl -X GET -b $authentication_cookie -H "Accept: application/json" "http://$3:$4/$5/admin?runMode=RESTART&accession=$exp_accession&type=makeexperimentprivate&autoDepends=false&op=schedule"
+                   fi
+               fi
+          else
+             err_msg="Updating private/public status of experiments on $3:$4/$5 unsuccessful: failed to find $exp_accession in AE2"
+             echo $err_msg >> $process_file.log
+             mailx -s "$err_msg" $6 < $process_file.log
```
olgamelnichuk, line 88: Not sure it is good idea to send mail for each failed experiment. In the worst case we get an email per atlas experiment, so about 5 000 emails..

scripts/cron/public_private_ae2_to_atlas.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
+# @author: rpetry
+# @date:   18 Oct 2011
+
+# This script serves to reflect in Atlas private/public experiment status switch on the AE2 side.
+
+if [ $# -eq 0 ]; then
+        echo "Usage: $0 ADMIN_USERNAME ADMIN_PASSWORD ATLAS_URL ATLAS_PORT ATLAS_ROOT ERROR_NOTIFICATION_EMAILADDRESS"
+        echo "e.g. $0 admin <pwd> lime 14032 gxa-load atlas-developers@ebi.ac.uk"
```
olgamelnichuk: Do we want to publish email address of our dev mailing list? May be we could use another example here?

rpetry: You're right - I'll just put a place holder here instead

scripts/cron/public_private_ae2_to_atlas.sh
rpetry, line 88: That is a very good point - I've added exit 1 - the processing should not continue in those cases

olgamelnichuk, line 88: Is it possible to have an experiment in atlas, but not in AE?

rpetry, line 88: in the unlikely event that the file retrieved from AE2 is incomplete - for whatever reason. Not being able to make a decision for a given Atlas experiment due to lack of data from AE2 would be an error situation.

PR #159, opened 2011-10-19
Trac 2849, Pretty property names moved to the properties themselves

alf239:
- DB changes are done
- migration scripts are ready
- Display names are supported in:
  - Experiment page
  - Tabs over charts
  - Gene table
  - Gene table search dropdown
  - Experiment design
  - Search results
  - Experiment popup in heatmap
  - Experiment index
  - Advanced search
  - EF dropdown
  - Search
    - EF in conditions lookup dropdown
    - Search results (list view)
  - Gene page's EF headers
- `LoaderDAO` generates accessions itself (ugly, but still)

Broken or missing functionality (comes soon after): 
- Display name editing via admin interface
- EF tabs in experiment list on gene page


atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/Assay.java
```diff
@@ -244,6 +247,18 @@ public class Assay {
         }
     }
 
+    public Collection<PropertyValue> getEffectiveValues(Property property) {
```
rpetry, line 264: Could you add javadoc on how you define an 'effective' value

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/AssayProperty.java
```diff
@@ -60,15 +60,7 @@ public final class AssayProperty {
...
     }
 
-    public AssayProperty(Assay assay, String name, String value, List<OntologyTerm> efoTerms) {
-        this.assaypvid = null; // TODO: 4alf: we must handle this on save
-        this.assay = assay;
-        propertyValue = new PropertyValue(null, new Property(null, name), value);
-        this.terms = new ArrayList<OntologyTerm>(efoTerms);
-    }
-
-    public AssayProperty(Long id, Assay assay, PropertyValue pv, List<OntologyTerm> efoTerms) {
```
rpetry, line 70: Why does Assay have a constructor with id in it, but AssayProperty doesn't. Is id in Assay just used for junit testing?

alf239, line 70: Historical reasons mostly; I'll take a look

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/Assay.java
alf239, line 264: here you are

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/Experiment.java
```diff
@@ -330,4 +330,15 @@ public class Experiment {
         }
         return null;
     }
+
+    public SortedSet<Property> getProperties() {
```
rpetry: Not sure I understand - this method is called in e.g. AtlasExperiment.getFactorValuesForEF().keys(). Going by the parent method name, shouldn't keys return just EF names, rather than EF and SC names?

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/PropertyValue.java
```diff
@@ -4,25 +4,34 @@ import org.hibernate.annotations.Cache;
...
     @Column(name = "NAME")
     private String value;
+    private String displayName;
 
     PropertyValue() {
     }
 
     public PropertyValue(Long id, Property definition, String value) {
+        if (definition == null)
+            throw new NullPointerException("Definition must be provided");
```
rpetry, line 31: Looking at calls to this constructor, definition should never be null. Should we not throw unexpected exception instead here?

alf239, line 31: It doesn't change anything, and should never happen, anyway. So I'm happy to change it to anything at all, as long as it's unchecked. Don't think NPE is bad or unsuitable, though.

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/Experiment.java
alf239: It's a bug actually, listed in outstanding tasks in http://bar.ebi.ac.uk:8080/trac/ticket/2849

Will review.

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/Sample.java
```diff
@@ -235,17 +209,13 @@ public class Sample {
         this.experiment = experiment;
     }
 
-    public void setChannel(String channel) {
-        this.channel = channel;
-    }
-
-    public boolean hasProperty(final PropertyValue propertyValue) {
+    private boolean hasNoProperty(final PropertyValue propertyValue) {
```
rpetry: Personally I find hasNoProperty method name confusing for two reasons - a. it sounds almost the same as another method: hasNoProperties() and at first sight sounds like both methods do the same b. Having No in the name confuses me more than than !hasProperty(). Could we not revert to the previous method definition?

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/Sample.java
```diff
@@ -258,12 +228,29 @@ public class Sample {
...
+        if (hasNoProperty(propertyValue)) {
+            addProperty(propertyValue, terms);
         } else {
-            SampleProperty sampleProperty = this.getProperty(propertyValue);
+            SampleProperty sampleProperty = getProperty(propertyValue);
             sampleProperty.setTerms(terms);
         }
     }
+
+    public Collection<PropertyValue> getEffectiveValues(Property property) {
```
rpetry: javadoc on the meaning of 'effective' in the new scheme of things

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/SampleProperty.java
```diff
@@ -22,21 +22,17 @@
```
rpetry, line 21: Same question as for AssayProperty - why does Sample have id in its constructor, but SampleProperty doesn't?

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/Sample.java
alf239: I don't think `hasNoProperties()` can create a problem here: one method has a parameter, another hasn't. On the inversion, there are different schools of though, so in order to save time I've rewritten the code.

atlas-updates/src/main/resources/db/migrations/201110171551-property-display-names.sql
```diff
@@ -0,0 +1,18 @@
+update a2_property set displayname='Clinical info' where name='clinical_information';
+update a2_property set displayname='Compound treatment' where name='compound';
+update a2_property set displayname='Genotype' where name='genotype';
```
alf239: Sources were:
1. `Curated.properties`
2. `select * from a2_config_property where name like 'factor.curatedname.%'`

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/SampleProperty.java
alf239, line 21: One I've cleaned up (unused constructor), and another was used somewhere. It wasn't a cleanup change, so I've only dropped what's obviously unused.

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/Sample.java
alf239: Renamed

alf239: Javadoc added, too.

atlas-web/src/main/java/ae3/service/structuredquery/AtlasEfvService.java
```diff
@@ -84,14 +87,16 @@ public class AtlasEfvService implements AutoCompleter, IndexBuilderEventHandler,
         this.propertyDAO = propertyDAO;
     }
 
-    public Set<String> getOptionsFactors() {
+    public Set<Property> getOptionsFactors() {
```
rpetry, line 90: Would you mind changing getOptionsFactors() to getOptionsIgnoredFactors(). The intended function is somewhat lost in the name otherwise.

atlas-updates/src/main/resources/db/migrations/201110171551-property-display-names.sql
alf239: By the way, leaving `update a2_property set displayname='Genotype' where name='genotype';` was a mistake: I was trying not to set display names where `StringUtil.prettify()` would do the same.

No need to remove, though—it's harmless.

atlas-web/src/main/java/uk/ac/ebi/gxa/properties/EfHeadersStorage.java
```diff
@@ -38,10 +40,10 @@ public class EfHeadersStorage implements Storage {
     private final static String PREFIX = "factor.curatedname.";
     private static final int SUFFIXPOS = PREFIX.length();
```
rpetry: no longer used

atlas-web/src/main/java/ae3/service/structuredquery/AtlasEfvService.java
alf239, line 90: I would actually: it returns all the factors _except_ ignored, so  `getOptionsIgnoredFactors()` would be even more confusing. I would be fine with `getOptionFactorsWithoutTheFactorsWeIgnore()`, but that looks too verbose and Monty Python-ish.

atlas-web/src/main/java/uk/ac/ebi/gxa/properties/EfHeadersStorage.java
alf239: bloody survivors! removed.

atlas-web/src/main/java/uk/ac/ebi/gxa/web/controller/GenePageExperiment.java
```diff
@@ -8,6 +9,7 @@ import uk.ac.ebi.microarray.atlas.model.Experiment;
 
 import java.util.Set;
 
+@JsonSerialize
```
rpetry, line 12: What does this annotation give us? And if needed, why doesn't GenePageGene have it too?

alf239, line 12: It shows that the class will be serialized with Jackson to avoid confusion with usage search. As of `GenePageGene`, didn't change it—not in scope.

atlas-web/src/main/webapp/WEB-INF/jsp/experimentpage/experiment.jsp
```diff
@@ -311,9 +312,9 @@
                             <select id="efvFilter" style="width:100%;">
                                 <option value="">All factor values</option>
                                 <c:forEach var="EF" items="${exp.experimentFactors}">
-                                    <optgroup label="${f:escapeXml(atlasProperties.curatedEfs[EF])}">
+                                    <optgroup label="${f:escapeXml(EF.displayName)}">
```
rpetry, line 315: Do we need to `f:escapeXml(EF.displayname)`? I didn't need to in [experiment-index.jsp](https://github.com/gxa/gxa/blob/develop/atlas-web/src/main/webapp/WEB-INF/jsp/experimentpage/experiment-index.jsp).

atlas-web/src/main/webapp/WEB-INF/jsp/includes/atlas-searchform.jsp
```diff
@@ -198,8 +198,8 @@
                         <option value="" selected="selected">Experimental factor</option>
                         <option value="">(any)</option>
                         <option value="efo">EFO</option>
-                        <c:forEach var="i" items="${atlasQueryService.experimentalFactorOptions}">
-                            <option value="${f:escapeXml(i)}">${f:escapeXml(atlasProperties.curatedEfs[i])}</option>
+                        <c:forEach var="factor" items="${atlasQueryService.experimentalFactorOptions}">
+                            <option value="${factor.name}">${f:escapeXml(factor.displayName)}</option>
```
rpetry, line 202: Do we need to `f:escapeXml`?

alf239, line 202: Imagine a pretty name like "Dose for 0 < t <= 5"—we'd better replace `<` with `&lt;`. 

I'm changing it to `<c:out value="" />` where appropriate (`c:out` escapes XML by default), but this one hasn't got there.

atlas-web/src/main/webapp/WEB-INF/jsp/experimentpage/experiment.jsp
alf239, line 315: We should always escape the values we put on the page unless we know they are sanitized. See also [OWASP Top 10 2010-A2-Cross-Site Scripting (XSS)](https://www.owasp.org/index.php/Top_10_2010-A2)

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/AssayProperty.java
alf239, line 70: Thanks for spotting—dropped the id

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/SampleProperty.java
alf239, line 21: Thanks for spotting—dropped the id

atlas-web/src/main/java/ae3/service/structuredquery/AtlasEfvService.java
rpetry, line 90: Ah sorry - my mistake - you're right - it's fine as it is

atlas-web/src/main/webapp/WEB-INF/jsp/experimentpage/experiment.jsp
alf239, line 315: By the way, here's the code from experiment-index.jsp:

``` jsp
                    <c:forEach var="factor" items="${experiment.experimentFactors}">
                        ${f:escapeXml(atlasProperties.curatedEfs[factor])}<br/>
                    </c:forEach>
```

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/Experiment.java
alf239: Addressed in 4693210

alf239:
On `breast cancer cell line`, it's a data problem; we must have properties sanitised in DB or else index will be useless. We may of course sanitize properties as we read those from DB—but that would mean we don't trust our own data, which would be an extremely weak position.

PR #163, opened 2011-10-25
Trac 2849: Loader sanitises names on the fly

alf239:
This is better to be reviewed/merged after #159


alf239:
Joined it with #159

PR #169, opened 2011-10-27
Tracking page state changes done with JavaScript

olgamelnichuk:
Implemented page state saving strategy:
- there is a global `atlas.pageState` objects, that handles all changes on the page done by components
- the global state is an object tree: state of a child component is saved in the state of its parent
- the global page state is serialised as name/value pairs and added as hash tag to the page url
- all components can register themselves in the `pageState` object, so that it could track their changes
- only root components should handle external hash change event (e.g. clicking on back/forward buttons)
- it is easy to turn off this feature by removing `atlas.PageState.init()` from a page, so all components can work with or without global page state

Minor change on gene page:
- 'Show all studies' link doesn't belong to pagination component any more;

This change automatically fixes [Trac 3002](http://bar.ebi.ac.uk:8080/trac/ticket/3002)
and makes step forward to fix [Trac 2617](http://bar.ebi.ac.uk:8080/trac/ticket/2617)


atlas-web/src/main/webapp/scripts/jquery.address-1.4.js
```diff
@@ -0,0 +1,666 @@
```
rpetry, line 0: Is this code actually ours or (original/adapted by you) from http://www.asual.com/jquery/address/?  If some or all of it is not ours, might be worth including an acknowledgement, no?

olgamelnichuk, line 0: Aah yes, damn Idea copyright plugin..

atlas-web/src/main/webapp/scripts/src/core/base64.js
```diff
@@ -0,0 +1,137 @@
```
rpetry, line 0: Does this script need copyright note also?

olgamelnichuk, line 0: No no.. this one is ours :)

PR #170, opened 2011-10-28
Never ever use quotes

alf239:
Oracle actually treats field and table names in a case-insensitive manner unless you use quotes around identifiers. If you create a table without quotes around the name, for example `CREATE MyTable...`, the resulting table name will be converted to upper case (i.e. `MYTABLE`) and will be treated in a case insensitive manner. `SELECT * from MYTABLE`, `SELECT * from MyTable`, `SELECT * from myTabLe` will all match `MYTABLE` (note the lack of quotes around the table name). Here is [a nice article](http://www.alberton.info/dbms_identifiers_and_case_sensitivity.html) on this issue that discusses this issue in more detail and compares databases.


alf239:
Nothing too fancy, just fixing old bugs — found by @olgamelnichuk

PR #178, opened 2011-11-07
Ticket #3175 - display links on gene page only if field.urlAware == 'true

rpetry:
Ticket #3175 - display links on gene page only if field.urlAware == 'true' (i.e. its underlying genproperty.link.<property> non empty)

### DIFF
--- a/atlas-web/src/main/webapp/WEB-INF/jsp/genepage/gene.jsp
+++ b/atlas-web/src/main/webapp/WEB-INF/jsp/genepage/gene.jsp
@@ -153,7 +153,7 @@ $(document).ready(function() {
                             <td class="propname">${f:escapeXml(field.name)}</td>
                             <td align="left">
                                 <c:choose>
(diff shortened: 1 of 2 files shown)


atlas-web/src/main/webapp/WEB-INF/jsp/genepage/gene.jsp
```diff
@@ -153,7 +153,7 @@ $(document).ready(function() {
                             <td class="propname">${f:escapeXml(field.name)}</td>
                             <td align="left">
                                 <c:choose>
-                                    <c:when test="${!empty field.urlAware}">
+                                    <c:when test="${field.urlAware == 'true'}">
```
olgamelnichuk: Wouldn't it be easier to write c:when test="${field.urlAware}" ?

rpetry: It would - modified - thanks

PR #183, opened 2011-11-08
Trac 3030: Dropping the "curated" status as unused

alf239:
http://bar.ebi.ac.uk:8080/trac/ticket/3030

Dropped the whole `curated` functionality except for the field itself: we're likely to switch back and forth between versions during development, so dropping a field from a table would do more harm than good.


rpetry:
It would be great to have migration to drop CUR_SetUnCurated.sql and CUR_SetCurated.sql - to avoid any confusion.

alf239:
These have nothing to do with this change; will drop, anyway.

alf239:
On a second thought, no, I won't. It should be a separate task, that is dropping all the curators' stored procedures and views. As far as I can see, `CUR_SetUnCurated` and `CUR_SetCurated` actually write a string to experiment log, that's it. It may or may not be useful, but it's not what created the original problem.

rpetry:
Ok. fair enough - we'll drop all CUR_\* stored procs when none of them are needed any more. Thanks.

PR #184, opened 2011-11-08
Ticket #3174 - Prevented re-direct loop between heatmap and gene page

rpetry:


rpetry:
A data fix will be implemented for this issue instead

PR #194, opened 2011-11-11
A script to summarize errors/exceptions from today's atlas log file

rpetry:

### DIFF
--- a/scripts/find_web_errors.sh
+++ b/scripts/find_web_errors.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+# @author: rpetry
+# @date:   11 Nov 2011
+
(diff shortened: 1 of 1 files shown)


scripts/find_web_errors.sh
```diff
@@ -0,0 +1,45 @@
...
+# This script is to be run on each of Atlas production tomcat machines
+
+if [ $# -eq 0 ]; then
+        echo "Usage: $0 NOTIFICATION_EMAILADDRESS"
+        exit;
+fi
+
+today="`eval date +%Y-%m-%d`"
+process_file="/tmp/find_web_errors.$today"
+log_dir="/ebi/www/jlive/servers/tc-fg-gxa/logs"
```
olgamelnichuk: Not sure it is a good idea to publish this path... can we just pass it as a parameter to the script?

rpetry: Good point - removed - thanks

PR #201, opened 2011-11-16
Trac3022

rpetry:
Implemented the following Curation API use cases:
- Retrieve all assay/sample properties by property name
- Retrieve all assay/sample properties by property name-value
- Retrieve all assays/samples that contain a property value mapped to an ontology term
- Retrieve all experiments that contain a given assay/sample
- Delete all assay properties containing a given factor
- Delete all assay properties containing a given factor value
- Delete all sample properties containing a given factor
- Delete all sample properties containing a given factor value
- Added junit tests to test the above


alf239:
As far as I can see from the [test cases](http://bar.ebi.ac.uk:8080/trac/wiki/ArrayExpressAtlas/CuratorsAPITestCases), we never use assays or samples per se. So it would be more reasonable to let curators retrieve experiments by property/term. That would also hide the implementation details, and reduce the public API.

atlas-web/src/main/java/uk/ac/ebi/gxa/web/controller/api/CurationApiController.java
```diff
@@ -95,6 +93,118 @@ public class CurationApiController extends AtlasViewController {
...
+            method = RequestMethod.GET)
+    @ResponseStatus(HttpStatus.OK)
+    public Collection<ApiExperiment> getExperimentsBySamplePropertyValue(@PathVariable("v") final ApiVersionType version,
+                                                           @PathVariable("propertyName") final String propertyName,
+                                                           @PathVariable("propertyValue") final String propertyValue,
+                                                           HttpServletResponse response) throws ResourceNotFoundException {
+        return curationService.getExperimentsBySamplePropertyValue(propertyName, propertyValue);
+    }
+
+    @RequestMapping(value = "/experiments//assays/ontologyterms/{ontologyTerm}",
```
alf239: duplicated slash

alf239:
On a code style: is there any reason to add `@param` doc-comments without any description? It doesn't add anything but clutter; IDEA keeps complaining, so I wonder why we keep those?

rpetry:
Re: on code style: there's no description if the variable names are self-descriptive. i could go round and either remove those @param fields or enter useless documentation, but I won't - given that we all use templates to create javadoc and we'd always need to remove @param fields that don't need documentation. I don't think this is practical.

atlas-web/src/main/java/uk/ac/ebi/gxa/web/controller/api/CurationApiController.java
rpetry: done - thanks

atlas-web/src/main/java/uk/ac/ebi/gxa/web/controller/api/CurationApiController.java
```diff
@@ -59,7 +57,7 @@ public class CurationApiController extends AtlasViewController {
         curationService.removePropertyValue(propertyName, propertyValue);
     }
 
-    @RequestMapping(value = "/experiments/assays/properties/{propertyName}/{oldPropertyValue}/{newPropertyValue}",
+    @RequestMapping(value = "/experiments/properties/{propertyName}/{oldPropertyValue}/{newPropertyValue}",
             method = RequestMethod.PUT)
     @ResponseStatus(HttpStatus.CREATED)
     public void replacePropertyValueInAssays(
```
alf239: So is it for assays or experiments?

rpetry: Experiments.

alf239: Could you please fix the method name?

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/api/ApiShallowExperiment.java
```diff
@@ -0,0 +1,75 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.Collection;
+
+import static com.google.common.collect.Collections2.transform;
+
+/**
+ * @author Robert Petryszak
+ *         <p/>
+ *         A minimal version of ApiExperiment
```
alf239: While it's not by any means a requirement, usually you start with a one-liner summarising a meaning of the class. After that goes the long description with optional examples, after that go all the @ tags.

Getting it other way around fools the tools and carries away the attention of the reader.

See also http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#format and http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#examples

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/api/ApiShallowExperiment.java
```diff
@@ -0,0 +1,75 @@
...
+            new Function<ApiSample, String>() {
+                @Override
+                public String apply(@Nonnull ApiSample input) {
+                    return input.getAccession();
+                }
+            };
+
+    private String accession;
+    private Collection<String> assayAccessions;
+    private Collection<String> sampleAccessions;
```
alf239: Why do you need this state? What's wrong with storing an `experiment` reference and getting the data on demand?

It's especially unclear in case of `private String accession`.

atlas-web/src/main/java/uk/ac/ebi/gxa/web/controller/api/CurationApiController.java
rpetry: Ah well spotted - fixed.

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/api/ApiShallowExperiment.java
rpetry: Aha - thanks for the references - I've fixed the order.

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/api/ApiShallowExperiment.java
rpetry: I guess the same could be said for ApiExperiment. Since some of the fields require transformation, it's done at construction time. Since we're already storing some fields separately from the Experiment reference, we decide to get all the fields we need out of Experiment and then release its reference for GC. The same thinking is applied in ApiShallowExperiment.

alf239: I fail to see the causation link between "require transformation" and "it's done at construction time".

The reference to GC doesn't seem to be relevant, either: as a matter of fact, you have session opened for a lifetime of the request, and session-level cache stores all the persistent objects unless you explicitly evict those.

PR #214, opened 2011-12-01
Ticket #3300 - Prevented experiments are being loaded with multiple efvs...

rpetry:
... per ef (in on assay) and multiple scvs per sc (in one sample).


atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/SourceStep.java
```diff
@@ -94,7 +94,6 @@ public class SourceStep {
                                         characteristicsAttribute + ") will be ignored"
                         );
                     }
-                }
```
alf239, line 97: code style: right now it's easy to understand what the change is about, even if you reformat the code; next accidental reformatting will add a change which will be irrelevant to work in progress

rpetry, line 97: Thanks for spotting that - fixed code alignment

PR #215, opened 2011-12-01
Ticket #3303 - MIgrated to limpopo 1.1

rpetry:


alf239:
It would be helpful to understand the rationale behind the migration.

rpetry:
So the message from Tony about 1.0.7 vs 1.1 is that:
- 1.0.7 is no longer supported
- 1.1:
  - should be faster
  - contains some bug fixes, object model simplifications
  - some old parallelisation issues (inherited from the original AE2 loader) have been eliminated.

The new version has been around for a few months now and is being used by AE2 - I don't see why we shouldn't upgrade on our side also.

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ParsingStep.java
```diff
@@ -51,9 +54,8 @@ public class ParsingStep {
     }
 
     public MAGETABInvestigation parse(URL idfFileLocation) throws AtlasLoaderException {
-        MAGETABInvestigation investigation = new MAGETABInvestigation();
+        MAGETABInvestigation investigation = null;
```
alf239: Why do we need this `null` here? If we parse IDF correctly, we can return proper `MAGETABInvestigation`; if we fail to parse, we throw an exception.

Please eliminate local state wherever you can: it only adds to the mess developers need to keep in their heads.

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ParsingStep.java
```diff
@@ -86,22 +88,25 @@ public class ParsingStep {
             }
         });
 
-        ExecutorService parseService = Executors.newFixedThreadPool(2);
-        ExecutorService idfService = Executors.newFixedThreadPool(2);
-        ExecutorService sdrfService = Executors.newFixedThreadPool(2);
```
alf239, line 91: These three lines are good news indeed!

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ParsingStep.java
```diff
@@ -86,22 +88,25 @@ public class ParsingStep {
...
-            throw new AtlasLoaderException("Parse error: " + e.getErrorItem().toString(), e);
-        } finally {
-            parseService.shutdownNow();
-            idfService.shutdownNow();
-            sdrfService.shutdownNow();
+            throw new AtlasLoaderException("Parse error: " + getParseErrors(e.getErrorItems()), e);
+
         }
 
         return investigation;
```
alf239: Generally, forget Wirth and avoid returning at the very end of the method: in Java, return at the end of the `try` block is usually much more readable. In Wirth's time, we had no exceptions.

Additionally, the more state you have, the more state one need to keep in one's head to comprehend the code. I'm sorry I'm picking on it here—it's just a good place to show how a variable can be eliminated.

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ParsingStep.java
```diff
@@ -86,22 +88,25 @@ public class ParsingStep {
...
+            investigation = parser.parse(idfFileLocation);
             log.info("Parsing finished");
         } catch (ParseException e) {
             // something went wrong - no objects have been created though
             log.error("There was a problem whilst trying to parse " + idfFileLocation, e);
-            throw new AtlasLoaderException("Parse error: " + e.getErrorItem().toString(), e);
-        } finally {
-            parseService.shutdownNow();
-            idfService.shutdownNow();
-            sdrfService.shutdownNow();
```
alf239, line 102: Three more lines of happiness!

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ParsingStep.java
```diff
@@ -86,22 +88,25 @@ public class ParsingStep {
...
+            throw new AtlasLoaderException("Parse error: " + getParseErrors(e.getErrorItems()), e);
+
```
alf239: When you'll be adjusting the code, drop this empty line please. Just æsthetics, nothing more.

alf239:
Mostly happy with the change (assuming you tested it and it works)—ther's just a bunch of minor comments to address.

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/service/AtlasMAGETABLoader.java
```diff
@@ -318,12 +318,12 @@ public class AtlasMAGETABLoader {
         // check that data is from RNASeq (comments: "Comment [ENA_RUN]"    "Comment [FASTQ_URI]" must be present)
         Collection<ScanNode> scanNodes = investigation.SDRF.lookupNodes(ScanNode.class);
```
olgamelnichuk: lookupNodes is deprecated:
- @deprecated since version 1.1, use {@link #getNodes(Class)} instead

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ParsingStep.java
rpetry: Agreed - fixed

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ParsingStep.java
rpetry: Ok - eliminated.

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ParsingStep.java
rpetry: done.

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/service/AtlasMAGETABLoader.java
rpetry: Well spotted - fixed, thanks.

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ParsingStep.java
```diff
@@ -51,9 +54,8 @@ public static String displayName() {
     }
 
     public MAGETABInvestigation parse(URL idfFileLocation) throws AtlasLoaderException {
-        MAGETABInvestigation investigation = new MAGETABInvestigation();
+        MAGETABInvestigation investigation;
```
olgamelnichuk: Could you please move the declaration of investigation variable to the try/catch block where it is actually used...

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ParsingStep.java
```diff
@@ -86,22 +88,23 @@ public void errorOccurred(ErrorItem item) {
...
 
-        return investigation;
+    public String getParseErrors(Collection<ErrorItem> errorItems) {
+        return on(',').join(transform(errorItems, new Function<ErrorItem, String>() {
+            @Override
+            public String apply(@Nonnull ErrorItem errorItem) {
+                return errorItem.reportString();
+            }
+        }));
     }
```
olgamelnichuk, line 108: This method is not used outside the ParsingStep class. Why it is public?

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/service/AtlasMAGETABLoader.java
olgamelnichuk: Here are all classes where lookupNodes is still used:
ArrayDataStep.java
AssayAndHybridizationStep.java
DerivedArrayDataMatrixStep.java
HTSArrayDataStep.java
SourceStep.java

olgamelnichuk:
The new version of limpopo parser returns non empty values only. I think this is good for us, as we can now eliminate code that creates "(empty)", "", " ", "  " values. The database also contains records for "(empty)", "", " ", "  " property values, and it would be nice to remove them also. 

Merging this changes probably will not brake atlas much, as we are filtering empty values all over the code. So if it is critical we could merge it asap, but I really like to see the code cleaned up and tested properly.

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/service/AtlasMAGETABLoader.java
rpetry: Done - thanks - sorry should have looked around myself.

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ParsingStep.java
rpetry: done

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ParsingStep.java
rpetry, line 108: changed to private

rpetry:
'The variable is still unused — may we inline it? Pretty please?'  - sure

PR #216, opened 2011-12-05
Trac 3030: dropping now-unused  column

alf239:
Dropped `curated` from `a2_experiment`


rpetry:
Should we not also drop curated from A2_Experiment ddl in Tables.sql?

alf239:
Mea culpa, fixed

PR #219, opened 2011-12-06
Ticket #3121 - Re-instated design element and gene mappings into to experiment loading process - they were missing before

rpetry:


alf239:
fine for a quick fix, we've a ticket for a proper follow-up

PR #220, opened 2011-12-07
Trac 2910 Cleaned up the save procedures; left readers intact

alf239:
- cleaned up the save procedures
- left readers intact
- added comments on our NetCDF peculiarities (could be done better)

### DIFF
--- a/atlas-data-storage/src/main/java/uk/ac/ebi/gxa/data/NetCDFDataCreator.java
+++ b/atlas-data-storage/src/main/java/uk/ac/ebi/gxa/data/NetCDFDataCreator.java
@@ -243,10 +243,16 @@ public class NetCDFDataCreator {
         );
         // !!!KLUDGE!!! Here be dragons
         //
-        // Actually, "GN" is a long array—and it is read as a long array
+        // Actually, "GN" is a long array&mdash;and it is read as a long array
+        //
+        // This is done this way since NetCDF v3 has no <code>LONG</code> type, so a long time ago we've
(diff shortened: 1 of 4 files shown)


atlas-data-storage/src/main/java/uk/ac/ebi/gxa/data/NetCDFDataCreator.java
```diff
@@ -241,6 +241,13 @@ private void create(NetcdfFileWriteable netCdf) throws IOException {
                 "DEacc", DataType.CHAR,
                 new Dimension[]{designElementDimension, designElementLenDimension}
         );
+        // !!!KLUDGE!!! Here be dragons
+        //
+        // Actually, "GN" is a long arrayÑand it is read as a long array
```
olgamelnichuk: "Ñ" ??

alf239: ouch!

alf239: Fixed — and improved the description; before that, it didn't explain **why** we couldn't simply use `LONG`

alf239:
Note: I hasn't created a proper unit test—thinking about it now. Still, it's harmless so far, and is already tested manually.

There's even a unit test which sort of tests the sequence—but, again, it's more like getting a fixed point rather than properly testing to crash.

PR #225, opened 2011-12-13
Ticket #3321 - modified to also pass to ExperimentViewController the displayed experiment factor when user clicks on a plot in an experiment pop-up off heatmap page

rpetry:


rpetry:
It's interesting that I seem to be able to achieve inclusion of the experiment factor in 'Show expression profile' link in the experiment pop-up _only_ when I set wro4j.tag.debug=true in wro4j-tag.properties.

PR #226, opened 2011-12-15
Trac 3215: fixed the on-page gene search

alf239:
As a bonus, `Burkitt's lymphoma` is coming soon


alf239:
The approach is now as follows: we use the very same `Predicate` to find genes in bitindex and `ExperimentParts` in the experiment. That allows us to eliminate a lot of repeated logic, and makes sure we never mix "all genes" with "nothing found" any more.

PR #229, opened 2011-12-20
Added annoations for GeneSigDB

nsklyar:
Refactoring to support annotations from files.


annotator/src/main/java/uk/ac/ebi/gxa/annotator/loader/AnnotationSourceConnection.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.Collection;
+import java.util.Set;
+
+/**
+ * User: nsklyar
+ * Date: 26/10/2011
+ */
+public abstract class AnnotationSourceConnection <T extends AnnotationSource> {
+
```
olgamelnichuk: why not to make it as Interface?

annotator/src/main/java/uk/ac/ebi/gxa/annotator/dao/AnnotationSourceDAO.java
```diff
@@ -51,13 +50,20 @@ public AnnotationSourceDAO(SessionFactory sessionFactory, JdbcTemplate atlasJdbc
...
     }
 
-    public void save(AnnotationSource object) {
-        object.setLoadDate(new Date());
-        template.save(object);
+    public void save(AnnotationSource annSrc) {
+        annSrc.setLoadDate(new Date());
+        //ToDo: Find better solution.This method call is needed to initialize "name" field, if it was not called before name=null.
+        // NS didn't manage to find how to force hibernate to access the field via getter.
+        annSrc.getName();
```
olgamelnichuk: why do we need to save the name into DB if we could always dynamically generate it?

nsklyar: the field is needed for the DB—to have a column with unique values in table `A2_ANNOTATIONSRC`. After moving biomart related fields in a separate table  table `ANNOTATIONSRC` didn't have any unique column (combination of columns) anymore, so `name` column was added.

annotator/src/main/java/uk/ac/ebi/gxa/annotator/loader/biomart/BioMartAnnotator.java
```diff
@@ -119,48 +106,45 @@ public void updateAnnotations(String annotationSrcId) {
 
             beDataWriter.writeBioEntities(data, listener);
             beDataWriter.writePropertyValues(data.getPropertyValues(), listener);
-            beDataWriter.writeBioEntityToPropertyValues(data, annSrc, listener);
+            beDataWriter.writeBioEntityToPropertyValues(data, annSrc, false, listener);
```
olgamelnichuk: Can we guarantee here that all three `beDataWriter.*` calls be run in the same transaction?

annotator/src/main/java/uk/ac/ebi/gxa/annotator/loader/biomart/BioMartConnection.java
```diff
@@ -107,7 +119,9 @@ public String getOnlineMartVersion() throws BioMartAccessException {
         return version;
     }
 
-    public Collection<String> validateAttributeNames(Set<String> properties) throws BioMartAccessException {
+    @Override
+    public Collection<String> validateAttributeNames(Set<String> properties) throws AnnotationSourceAccessException {
+        connect();
```
olgamelnichuk: Instead of checking `if (connected == "true")` in every method, it would be easier to create `BioMartConnection` with already pre-fetched registry info; just use `static` method instead of constructor: `BioMartConnection.create(bioMartSource)`...

annotator/src/main/java/uk/ac/ebi/gxa/annotator/loader/AnnotationProcessor.java
```diff
@@ -0,0 +1,80 @@
...
+            } catch (NumberFormatException e) {
+                throw LogUtil.createUnexpected("Cannot fetch Annotation Source. Wrong ID ", e);
+            }
+        }
+        return annSrc;
+    }
+
+    private Annotator getAnnotator(String id) {
+        final AnnotationSource annotationSource = fetchAnnSrcById(id);
+        final AnnotationSourceType annSrcType = AnnotationSourceType.getByClass(annotationSource.getClass());
```
olgamelnichuk: why not to return the type by the annotationsource instance itself?
Probably bad idea.. 
or at least hide `annotationSource.getClass()`: e.g. `AnnotationSourceType.annTypeOf(annotationSource)`..

annotator/src/main/java/uk/ac/ebi/gxa/annotator/model/AnnotationSource.java
```diff
@@ -62,26 +63,68 @@
...
         this.software = software;
-        this.organism = organism;
     }
 
     public Long getAnnotationSrcId() {
         return annotationSrcId;
     }
 
+    @Column(name = "name")
+    public String getName() {
```
alf239, line 105: Having a getter with a side effect, _especially_ a side effect we rely on, is a very fragile idea: it's hard to spot, easy to forget, and super-easy to break later on. Why not create the name as we create a source?

annotator/src/test/java/uk/ac/ebi/gxa/annotator/dao/AnnotationSourceDAOTest.java
```diff
@@ -93,17 +112,32 @@ public void testSave() throws Exception {
 
     @Test
     public void testGetById() {
+        AnnotationSource annotationSource = annSrcDAO.getById(1000, BioMartAnnotationSource.class);
+        Assert.assertNotNull(annotationSource);
```
olgamelnichuk: Can we also be sure here that annotationSource is instance of BioMartAnnotationSource class?
I mean:

```
BioMartAnnotationSource annotationSource = annSrcDAO.getById(1000, BioMartAnnotationSource.class);
Assert.assertNotNull(annotationSource);
```

annotator/src/main/java/uk/ac/ebi/gxa/annotator/loader/AnnotationProcessor.java
nsklyar: makes sence, done

annotator/src/main/java/uk/ac/ebi/gxa/annotator/loader/AnnotationSourceConnection.java
nsklyar: changed

annotator/src/main/java/uk/ac/ebi/gxa/annotator/loader/biomart/BioMartAnnotator.java
nsklyar: they are running in different transactions, but it a "feature" :) The steps shouldbe done in order, but they are independent.

annotator/src/main/java/uk/ac/ebi/gxa/annotator/model/AnnotationSource.java
nsklyar, line 105: I've added call createName() into constructor, and remove from getter, but it's still easy to forget to call create name in new classes.

annotator/src/test/java/uk/ac/ebi/gxa/annotator/dao/AnnotationSourceDAOTest.java
nsklyar: data are taken fom our tast DB where annotation source with id 1000 is BioMartAnnotationSource

annotator/src/main/java/uk/ac/ebi/gxa/annotator/loader/biomart/BioMartConnection.java
nsklyar: fixed

PR #236, opened 2012-01-06
Ticket #3334

rpetry:
Fixed the logic to automatically move to using processed files for non-Affy experiments and when no raw files exist; otherwise try to process raw files and report an error on any failure.

### DIFF
--- a/atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ArrayDataStep.java
+++ b/atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ArrayDataStep.java
@@ -238,15 +238,13 @@ public class ArrayDataStep {
                                 if (localZipFile != null && !localZipFile.delete()) {
                                     log.error("Cannot delete " + localZipFile.getAbsolutePath());
                                 }
-                                log.error("IOException is thrown: " + e.getMessage());
-                                throw new AtlasLoaderException("Error occurred while retrieving raw data files from ArrayExpress ftp site" + USE_PROCCESSED_FILES);
+                                throw new AtlasLoaderException("Error occurred while retrieving raw data files from ArrayExpress ftp site" + USE_PROCCESSED_FILES, e);
                             }
                         }
                         try {
                             extractZip(localZipFile, adData.dataDir);
                         } catch (IOException e) {
-                            log.error("IOException is thrown: " + e.getMessage());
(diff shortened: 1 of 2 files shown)


atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/steps/ArrayDataStep.java
```diff
@@ -234,20 +238,16 @@ public void readArrayData(@Nonnull AtlasComputeService computeService, MAGETABIn
...
+                                throw new AtlasLoaderException("Error: '"+e.getMessage()+"' occurred while retrieving raw data files from ArrayExpress ftp site" + USE_PROCCESSED_FILES);
                             }
                         }
-
                         try {
                             extractZip(localZipFile, adData.dataDir);
                         } catch (IOException e) {
-                            log.warn("IOException is thrown: " + e.getMessage());
-                            throw new AtlasLoaderException("Error occurred while retrieving raw data files from ArrayExpress ftp site" + USE_PROCCESSED_FILES);
+                            throw new AtlasLoaderException("Error: '"+e.getMessage()+"' occurred while retrieving raw data files from ArrayExpress ftp site" + USE_PROCCESSED_FILES);
```
olgamelnichuk: Why not to just wrap exception "e" here: AtlasLoaderException(msg, e)?

rpetry: I was hoping to provide a more specific/meaningful message - as this will be presented to the curators loading the experiment.  I don't know how informative the IOException message would be.

olgamelnichuk: Than at least we should log it; to help ourselves to investigate the problem in the future

nsklyar: as far as I understand Olga's comment is not really addressed ... we can either log an error (as it used to be), or wrap it like: 
 throw new AtlasLoaderException("Error: occurred while retrieving raw data files from ArrayExpress ftp site" + USE_PROCCESSED_FILES, e);

I think inserting e.getMessage of IOException into your message might not be very meaningful for curators anyway.

rpetry: Done

PR #237, opened 2012-01-06
Playing with clover

alf239:
Finally got clover doing what we need. Until we upgrade Bamboo I suggest we stick to this state: we have proper reporting, and don't need much more just now.


alf239:
Currently, http://bar.ebi.ac.uk:8085/browse/ATLAS-METRICS runs off the `clover` branch, so feel free to check the results there.

alf239:
Pretty pictures if you like those: http://bar.ebi.ac.uk:8085/browse/ATLAS-METRICS-426/artifact/Clover/index.html

PR #238, opened 2012-01-17
Trac 3324: flipped statistics in some RNAsec experiments

alf239:
http://bar.ebi.ac.uk:8080/trac/ticket/3324

### DIFF
--- a/atlas-analytics/src/main/resources/R/analytics.R
+++ b/atlas-analytics/src/main/resources/R/analytics.R
@@ -118,11 +118,20 @@ read.atlas.nc <<-
 
 ### Log function that does not return NAs
 log2.safe <-
-  function(x,log = T) {
-    if (!log)
+  function(x) {
+    # Assumption is, expression levels are > 0 unless someone has _already_ applied log.
+    # If the levels are < 1000, we're fine without log: moreover, that may as well mean
+    # all the expression levels were > 1 (why not?), and someone has already applied log.
+    # If the values are more than 1000, then log definitely was not used before, as 2^1000
+    # is just way too much for expression level value.
+    if (min(x, na.rm = TRUE) < 0 || max(x, na.rm = TRUE) < 1000)
       return(x)
 
-    tmp = log2(x - min(x) + 1e-10)
+    print("Taking log2 of the expression matrix")
+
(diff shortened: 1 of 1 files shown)


atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -122,7 +122,7 @@ log2.safe <-
     if (!log)
       return(x)
 
-    tmp = log2(x)
+    tmp = log2(x - min(x) + 1e-10)
```
rpetry: Why is `min(x)` needed?

alf239: It's an artefact from the previous lives of this code: before, we might as well have negative values passed here, which would result in `log(-1) == NaN`, `is.finite(NaN) == FALSE`, and the same flipping trick as we had with zero.

Whether or not it is necessary right now is a tough call. We look for `max(exprs, na.rm = TRUE) > 1000 && min(exprs, na.rm = TRUE) >= 0` to decide on the need for logarithm, and it rules the `x < 0` option out. Still, we do that in a _different_ function, so following the principle of separating concerns, we should not rely on that.

What's more interesting is the case where all the expression levels are high, e.g. all are in thousands:

```
> log(c(1000.1, 1000.2, 1000.3, 1000.4))
[1] 6.907855 6.907955 6.908055 6.908155
> log(c(0.1, 0.2, 0.3, 0.4))
[1] -2.3025851 -1.6094379 -1.2039728 -0.9162907
```

Initially (seemingly) small difference becomes (seemingly) more significant due to the scaling. Whether it's a problem or not, I don't know. I don't think it is,  but then I haven't done any due diligence on that side.  We can keep it on a safe side checking that all non-NaN values in `x` are greater or equal to zero.

rpetry: Yes, because of the potential of the scaling problem, I'd prefer not to use `- min(x)` and just check inside `log2.safe` that all non-NaN values in `x` are greater or equal to zero. Also, would you mind commenting in code why `1e-10` is used ?

alf239: Here you are

PR #239, opened 2012-01-17
Trac 3204: Pretty EF names for experiment list on the gene page

alf239:


alf239:
My apologies for the @pashky-style map: it was the most concise way to express the conversion. 

Given that it's only used once, I see no need for a special class—yet.

PR #247, opened 2012-02-01
Trac 3150

alf239:
Eliminated most of the parallel structures getting from NetCDFs to the rest of the code.

The central idea of the change is the `StatisticsCursor` class enabling navigation over NetCDF without mentioning the indices—thus, all the index-related functionality along with handling of parallel structures is hidden in the cursor.

The only part where NetCDF internals are effectively exposed is `ExperimentPlot` and its JS counterpart: I simply have no time to change structures here, thus leaving it to you.

As of unit tests, started covering, it's just that I'd like to get feedback ASAP and not postpone the fixes.


atlas-data-storage/src/main/java/uk/ac/ebi/gxa/data/ExperimentWithData.java
```diff
@@ -214,25 +221,18 @@ public ExpressionStatistics getExpressionStatistics(ArrayDesign arrayDesign, int
...
-                        uniqueEFV.getKey(),
-                        uniqueEFV.getValue(),
-                        t[efIndex],
-                        p[efIndex]
-                ));
-            }
+
+        final Predicate<Pair<String, String>> efvPredicate;
+        if (efName == null)
+            efvPredicate = alwaysTrue();
```
alf239: `ANY_EFV` would be better. Not a big deal, but still.

atlas-model/src/main/java/uk/ac/ebi/microarray/atlas/model/ExpressionAnalysis.java
```diff
@@ -22,133 +22,63 @@
...
 
 /**
  * @author Tony Burdett
+ * @deprecated use {@link DesignElementStatistics instead}
  */
-public class ExpressionAnalysis implements Serializable, Comparable<ExpressionAnalysis> {
-    public static final long serialVersionUID = -6759797835522535043L;
-
+@Deprecated
+public class ExpressionAnalysis implements DesignElementStatistics {
```
alf239, line 34: Can be eliminated in favor of  `DesignElementStatistics`

atlas-data-storage/src/main/java/uk/ac/ebi/gxa/data/ExperimentWithData.java
alf239: addressed

PR #248, opened 2012-02-06
Ticket 2958 - remove empty property values from Atlas

rpetry:


atlas-updates/src/main/resources/db/migrations/201201060957-drop-empty-propertyvalues.sql
```diff
@@ -0,0 +1 @@
+delete from A2_propertyvalue where regexp_like(name, '^[ ]+$') or name = '(empty)';
```
olgamelnichuk, line 1: What about "" values? This statement doesn't remove them.

atlas-updates/src/main/resources/db/migrations/201201061001-trim-propertyvalues.sql
```diff
@@ -0,0 +1,20 @@
+-- First trim all property values that don't already exist in A2_propertyvalue in trimmed form
+update A2_propertyvalue pv set pv.name = trim(pv.name) where length(trim(pv.name)) <> length(pv.name)
+and not exists (select 1 from A2_propertyvalue where name = trim(pv.name) and propertyid = pv.propertyid);
+
```
olgamelnichuk, line 4: In case all values for a given propertyid are not trimmed e.g. "name ", "name    ", .. constraint violation will be thrown. Wondering if this error can appear on the production instance.

atlas-updates/src/main/resources/db/migrations/201201061001-trim-propertyvalues.sql
```diff
@@ -0,0 +1,20 @@
...
+where apv.propertyvalueid in (select propertyvalueid from A2_propertyvalue where length(trim(name)) <> length(name));
+
+update A2_samplepv apv
+set apv.propertyvalueid = (
+  select pv1.propertyvalueid from A2_propertyvalue pv1, A2_propertyvalue pv2
+  where pv1.name = trim(pv2.name) and pv2.propertyvalueid = apv.propertyvalueid and pv2.propertyid = pv1.propertyid)
+where apv.propertyvalueid in (select propertyvalueid from A2_propertyvalue where length(trim(name)) <> length(name));
+
+-- Finally remove now obsolete non-trimmed property values (from A2_propertyvalue and via FK constraints from A2_assaypv and A2_samplepv)
+delete from A2_propertyvalue pv where length(trim(name)) <> length(name) and exists (select name from A2_propertyvalue where name = trim(pv.name));
```
olgamelnichuk: Very scary 'deletion' line... I would prefer something like this

```
...and not exists (select 1 from a2_assaypv a, a2_samplepv s where a.propertyvalueid = pv.propertyvalueid or s.propertyvalueid = pv.propertyvalueid);
```

.. but need to check how fast it is

olgamelnichuk:
Can we be sure that the new loaded property values are trimmed?

atlas-updates/src/main/resources/db/migrations/201201060957-drop-empty-propertyvalues.sql
rpetry, line 1: in atlasraw2:
select count(*) from A2_propertyvalue where name = ''; --> 0 rows

atlas-updates/src/main/resources/db/migrations/201201061001-trim-propertyvalues.sql
rpetry, line 4: Good point - I had in fact tried this statement in atlasraw2 and then rolled it back - no errors were thrown.

atlas-updates/src/main/resources/db/migrations/201201061001-trim-propertyvalues.sql
rpetry: Scary, but logically correct. I've added the safety net you suggested - and confirmed that the same amount of rows is deleted with or without the safety net.

rpetry:
Yes property values are being trimmed at load time - please see changes in SourceStep and AssayAndHybridizationStep classes

PR #249, opened 2012-02-08
Tests - cleaning up R service based tests

olgamelnichuk:
- merged AtlasRFactory tests into a parameterised one RFactoryTest; to minimize code duplication
- moved rcommand utility classes and their tests into proper package: uk.ac.ebi.gxa.R.compute
- eliminated R.properties, as it only used in atlas-test/LoaderDriver which is not supported any more
- atlas-web module doesn't depend on useless atlas-test any more
- added ability to run unit tests with biocep.killused=false method, if the test calculations doesn't affect R server environment much
- moved biocep.properties back to test package as it is needed only for tests
- extracted biocep.defaultpool property to the maven settings.xml; so it is your choice now to use ATLAS_8G or ATLAS_32G
- added ability to retry to borrow biocep rservices for 5 times with interval of 20 seconds.. in case when all workers are busy and the pool is currently expanding..


atlas-analytics/src/main/java/uk/ac/ebi/gxa/R/AtlasRFactoryBuilder.java
```diff
@@ -120,150 +83,18 @@ public AtlasRFactory buildAtlasRFactory(RType rType) throws InstantiationExcepti
...
-                        Properties biocepProperties = new Properties();
-                        biocepProperties.load(getClass().getClassLoader().getResourceAsStream("biocep.properties"));
-                        setBiocepSystemProperties(biocepProperties);
-                    }
-                    catch (IOException e) {
-                        throw new InstantiationException("Cannot initialize AtlasRFactory for " + rType +
-                                " with null properties and without a biocep.properties file on the classpath");
-                    }
+                try {
+                    bicepPropertiesSetup(properties);
```
rpetry: typo

rpetry:
Great work - thanks Olga!

PR #251, opened 2012-02-10
Ticket 3117 - Eliminated isactive column from A2_software table

rpetry:


alf239:
Bravo maestro!

PR #256, opened 2012-02-20
Trac3356 - merging units into factor values; merging compounds with units

rpetry:


atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/service/PropertyValueMergeService.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /**
+     * @param term
+     * @return true if term can be found in EFO; false otherwise
+     */
+    private boolean isEfoTerm(String term) {
+        return efo.searchTerm(term).isEmpty();
+    }
+}
```
nsklyar, line 164: Probably it could be useful to have a unit test for this class.

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/service/PropertyValueMergeService.java
```diff
@@ -0,0 +1,159 @@
...
+                continue; // We don't load empty factor values
+
+            if (COMPOUND.equalsIgnoreCase(ef))
+                compoundFactorValue = efv;
+            else if (DOSE.equalsIgnoreCase(ef))
+                doseFactorValue = efv;
+
+            if (COMPOUND.equalsIgnoreCase(ef) || DOSE.equalsIgnoreCase(ef)) {
+                if (!Strings.isNullOrEmpty(compoundFactorValue) && !Strings.isNullOrEmpty(doseFactorValue)) {
+                    ef = COMPOUND;
```
nsklyar, line 128: do we need this operation?

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/service/PropertyValueMergeService.java
```diff
@@ -0,0 +1,159 @@
...
+            }
+            factorValues.add(Pair.create(ef, efv));
+        }
+
+        if (!Strings.isNullOrEmpty(compoundFactorValue) && Strings.isNullOrEmpty(doseFactorValue)) {
+            factorValues.add(Pair.create(COMPOUND, compoundFactorValue));
+            log.warn("Adding " + COMPOUND + " : " + compoundFactorValue + " to assay with no corresponding value for factor: " + DOSE);
+        } else if (!Strings.isNullOrEmpty(doseFactorValue) && Strings.isNullOrEmpty(compoundFactorValue)) {
+            throw new AtlasLoaderException(DOSE + " : " + doseFactorValue + " has no corresponding value for factor: " + COMPOUND);
+        }
```
nsklyar, line 148: is it right that  factor "dose" cannot exist without "compound"? it might be for example dose of radiation.

rpetry, line 148: Yes, as agreed with curators - if dose exists and compound doesn't, we fail the experiment load. They have committed to load into Atlas only experiments with both compound and dose, or just with compound - if dose is not available. If there indeed exist experiments with dose and radiation - then your question is for the curators. But the code is implementing the contract agreed with the curators as of now.

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/service/PropertyValueMergeService.java
rpetry, line 128: If I understand your question correctly - we do because we can only merge if both compoundFactorValue and doseFactorValue are present. Let's discuss if this is unclear or I got the wrong end of the stick.

atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/service/PropertyValueMergeService.java
rpetry, line 164: Both public methods in this class are now tested in TestAtlasMAGETABLoader

PR #257, opened 2012-02-20
Added "mappingsApplied" column for Annotation Sources in admin interface

nsklyar:


sql/Schema/Tables.sql
```diff
@@ -1671,6 +1671,7 @@ CREATE TABLE A2_ANNOTATIONSRC(
   , annsrctype VARCHAR2(255) CONSTRAINT NN_ANNOTATIONSRC_ASTYPE NOT NULL
   , LOADDATE DATE
   , isApplied VARCHAR2(1) DEFAULT 'F'
```
rpetry: Could we perhaps rename isApplied with annotationsApplied? Otherwise the two columns together are a bit confusing.

nsklyar: done

PR #268, opened 2012-03-07
Show user friendly validation messages when saving annotation source

nsklyar:


annotator/src/main/java/uk/ac/ebi/gxa/annotator/validation/ValidationReportBuilder.java
```diff
@@ -58,4 +63,20 @@ public String getSummary(String headMessage, String separator) {
         }
         return "";
     }
+
+    public String  getSummary(String messageIfValid, String headMessage, String separator) {
+        if (!isEmpty()) {
+            StringBuilder sb = new StringBuilder();
+            sb.append(headMessage);
+            sb.append(": ");
+            for (String message : messages) {
```
rpetry: A minor thing, but could you not use Joiner.on(separator).join(messages) ?

PR #276, opened 2012-03-13
Ticket 3431 - Replace multiple spaces and underscores with their single equivalents in all factor and sample characteristic values at experiment load time

rpetry:
@olgamelnichuk, @nsklyar - could you review please

### DIFF
--- a/atlas-utils/src/main/java/uk/ac/ebi/gxa/utils/StringUtil.java
+++ b/atlas-utils/src/main/java/uk/ac/ebi/gxa/utils/StringUtil.java
@@ -72,4 +72,11 @@ public final class StringUtil {
     public static String prettify(String s) {
         return s.replaceAll("_", " ");
     }
+
+    public static String removeSeparatorDuplicates(String s) {
+        s = s.replaceAll("( )+", " ");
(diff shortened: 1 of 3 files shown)


atlas-utils/src/main/java/uk/ac/ebi/gxa/utils/StringUtil.java
```diff
@@ -72,4 +72,11 @@ public static String limitedJoin(Collection<String> strings, int num, String sep
...
         return s.replaceAll("_", " ");
     }
+
+    public static String removeSeparatorDuplicates(String s) {
+        s = s.replaceAll("( )+", " ");
+        s = s.replaceAll("(_)+", "_");
+        return s.trim();
+
+    }
 }
```
nsklyar, line 80: probably it's just a metter of preference but can be shorter like:

``` java
 return s.replaceAll("( )+", " ")
                .replaceAll("(_)+", "_")
                .trim();
```

rpetry, line 80: Am happy to go with your preference - fixed.

PR #284, opened 2012-03-27
Trac3456

rpetry:
- Split experiments queries into separate UP and DOWN expression ones
- Modified to fetch experiments for a given expression type only if the counter reports any for that expression type. 

This will both prevent nonde expression experiments being output if the user hadn't queried for them, but will also make the API queries faster - but avoiding to unnecessarily query for experiments if we already know that none can be found.

@nsklyar - could you review please

### DIFF
--- a/atlas-web/src/main/java/uk/ac/ebi/gxa/requesthandlers/api/result/HeatmapResultAdapter.java
+++ b/atlas-web/src/main/java/uk/ac/ebi/gxa/requesthandlers/api/result/HeatmapResultAdapter.java
@@ -154,15 +154,8 @@ public class HeatmapResultAdapter implements ApiQueryResults<HeatmapResultAdapte
             }
 
             Iterator<ExperimentResult> expiter() {
-                EfvAttribute attr = new EfvAttribute(efefv.getEf(), efefv.getEfv());
-                List<ExperimentResult> allExps = Lists.newArrayList();
-                if (getUpExperiments() > 0)
-                    allExps.addAll(atlasStatisticsQueryService.getExperimentsSortedByPvalueTRank(row.getGene().getGeneId(), attr, -1, -1, StatisticsType.UP));
-                if (getDownExperiments() > 0)
-                    allExps.addAll(atlasStatisticsQueryService.getExperimentsSortedByPvalueTRank(row.getGene().getGeneId(), attr, -1, -1, StatisticsType.DOWN));
-                if (getNonDEExperiments() > 0)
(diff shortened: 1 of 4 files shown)


atlas-web/src/main/java/uk/ac/ebi/gxa/requesthandlers/api/result/HeatmapResultAdapter.java
```diff
@@ -178,8 +184,13 @@ public String getEfoId() {
...
+                List<ExperimentResult> allExps = Lists.newArrayList();
+                if (getUpExperiments() > 0)
+                    allExps.addAll(atlasStatisticsQueryService.getExperimentsSortedByPvalueTRank(row.getGene().getGeneId(), attr, -1, -1, StatisticsType.UP));
+                if (getDownExperiments() > 0)
+                    allExps.addAll(atlasStatisticsQueryService.getExperimentsSortedByPvalueTRank(row.getGene().getGeneId(), attr, -1, -1, StatisticsType.DOWN));
+                if (getNonDEExperiments() > 0)
+                    allExps.addAll(toNonDEResults(atlasStatisticsQueryService.getScoringExperimentsForBioEntityAndAttribute(row.getGene().getGeneId(), attr, NON_D_E)));
                 return allExps.iterator();
             }
         }
```
nsklyar, line 190: it's possible to avoid code duplication if you create method

```
protected Iterator<ExperimentResult> createExpIter(Attribute attr) {
                List<ExperimentResult> allExps = Lists.newArrayList();
                if (getUpExperiments() > 0)
                    allExps.addAll(atlasStatisticsQueryService.getExperimentsSortedByPvalueTRank(row.getGene().getGeneId(), attr, -1, -1, StatisticsType.UP));
                if (getDownExperiments() > 0)
                    allExps.addAll(atlasStatisticsQueryService.getExperimentsSortedByPvalueTRank(row.getGene().getGeneId(), attr, -1, -1, StatisticsType.DOWN));
                if (getNonDEExperiments() > 0)
                    allExps.addAll(toNonDEResults(atlasStatisticsQueryService.getScoringExperimentsForBioEntityAndAttribute(row.getGene().getGeneId(), attr, NON_D_E)));
                return allExps.iterator();
            }
```

in Expression class

rpetry, line 190: Done

PR #294, opened 2012-04-16
Ticket 3467 - Simplification of the output of experimental condition autocomplete 

rpetry:
Before this change:

1.In the case ef1:val, ef2:val, autocomplete shows both ef1:val and ef2:val - with separate counts for each of them  
2. In the case of efo:val, ef1:val, autocomplete shows both efo:val, ef1:val - with separate counts for each of them     

Implemented in this change:
Re 1: val is shown once, with no factors and the cumulative count across ef1 and ef2
Re 2: val is shown once, with efo:val's count only and excluding (EFO_...) identifier. Users do not need to see EFO_... identifiers.

@nsklyar - could you review please?


atlas-web/src/main/java/ae3/service/structuredquery/AutoCompleteResult.java
```diff
@@ -0,0 +1,51 @@
...
+    private final Multimap<String, AutoCompleteItem> autoCompleteCandidates = Multimaps.newListMultimap(
+            Maps.<String, Collection<AutoCompleteItem>>newHashMap(),
+            new Supplier<List<AutoCompleteItem>>() {
+                @Override
+                public List<AutoCompleteItem> get() {
+                    return Lists.newArrayList();
+                }
+            }
+    );
+
```
nsklyar, line 27: Here we probably can have just a map, since we always take only one first values from the list if more then one is present (line:63)

rpetry, line 27: So we only take the first item for 'efoefv' type of autocomplete, whereas for 'efv' or 'efo' autocomplete we take all items - as before this change.  Hence the need to use  newListMultimap to cover all three cases.

atlas-web/src/main/java/ae3/service/structuredquery/AtlasEfvService.java
```diff
@@ -195,25 +207,20 @@ private boolean isProhibited(Property property) {
...
+    private Multiset<String> treeAutocomplete(Collection<Property> properties, final @Nonnull String prefix, final int limit) {
+        final Multiset<String> result = HashMultiset.create();
 
         for (final Property property : properties) {
             PrefixNode root = treeGetOrLoad(property.getName());
             if (root != null) {
                 root.walk(prefix, 0, "", new PrefixNode.WalkResult() {
                     public void put(String name, int count) {
-                        result.add(
-                                new EfvAutoCompleteItem(
```
nsklyar, line 207: class EfvAutoCompleteItemis not used anymore

atlas-web/src/main/java/ae3/service/structuredquery/AutoCompleteResult.java
nsklyar, line 27: ok, I see it's used for gene autocompletion as well, where multiple values are possible.

Then it's fine with me to merge this change (after addressing another comment).

atlas-web/src/main/java/ae3/service/structuredquery/AtlasEfvService.java
rpetry, line 207: Removed

PR #296, opened 2012-04-18
Trac3468 - Hide from user factor selection on the advanced search interface (it confuses them)

rpetry:
@nsklyar - could you review please?


atlas-web/src/main/webapp/scripts/atlas-searchform.js
```diff
@@ -534,15 +534,6 @@ var atlas = atlas || {};
         function addExpFactor(factor, expression, minexps, values) {
             var selopt = $('#factors').get(0).options;
```
nsklyar: var selopt is unused

atlas-web/src/main/webapp/WEB-INF/jsp/includes/atlas-searchform.jsp
```diff
@@ -167,12 +167,8 @@
                 </td>
                 <td>
                     <select id="factors" style="width:320px">
-                        <option value="" selected="selected">Experimental factor</option>
+                        <option value="" selected="selected">Condition</option>
                         <option value="">(any)</option>
-                        <option value="efo">EFO</option>
-                        <c:forEach var="factor" items="${atlasQueryService.experimentalFactorOptions}">
```
nsklyar, line 173: Does it make sence to leave a choice between "Experimental factor" and "any"? what does "any" mean now?

atlas-web/src/main/webapp/scripts/atlas-searchform.js
rpetry: Removed

atlas-web/src/main/webapp/WEB-INF/jsp/includes/atlas-searchform.jsp
rpetry, line 173: 'Experimental Factor' dropdown is now a 'Condition', but I agree that this is not ideal. I think that the advanced interface will need an overhaul, but this will have to wait till we have clear requirements for the baseline vs differential Atlas. The change in this request is a minimal one that eliminates user's confusion with having to select factors before selecting values.

PR #302, opened 2012-04-30
Ticket 3478 - Made expression type work for gene-only queries

rpetry:
@nskyar - please review

### DIFF
--- a/atlas-web/src/main/webapp/scripts/atlas-searchform.js
+++ b/atlas-web/src/main/webapp/scripts/atlas-searchform.js
@@ -213,8 +213,6 @@ var atlas = atlas || {};
         //   minExperiments - optional (default is 1)
         // }
         this.addCondition = function (condition) {
-            var v = (condition.value = trim(condition.value || ""));
-            var f = (condition.factor || "");
             q.conditions.push(condition);
         };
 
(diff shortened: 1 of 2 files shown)


atlas-web/src/main/webapp/scripts/atlas-searchform.js
```diff
@@ -213,9 +213,7 @@ var atlas = atlas || {};
         this.addCondition = function (condition) {
             var v = (condition.value = trim(condition.value || ""));
             var f = (condition.factor || "");
-            if (v.length || f.length) {
-                q.conditions.push(condition);
-            }
```
nsklyar, line 218: this check should not be removed! Expected behaviour:  If no value or factor is given, then warning message shown. if this condition removed that doesn't happen.

rpetry, line 218: Re-instated the expected behaviour - see my change. But we need to let an empty experimental condition through if a gene condition was specified - this is so that we don't loose the user's expression type selection

nsklyar, line 218: probably need to remove unused variables v and f

rpetry, line 218: Done

PR #333, opened 2012-06-13
Trac3512 - Comments added (on behalf of @mkeays)

rpetry:


atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
-       colnames(efv) = ef
-       rownames(efv) = as
-       efv = data.frame(efv)
-       print(paste("Read in EFV:", nrow(efv), "x", ncol(efv)))
-    }
 
-    if ("SC" %in% names(nc$dim)) {
-      sc = get.var.ncdf(nc, "SC")
-      scv = get.var.ncdf(nc, "SCV")
+# commented excessively in June2012 by mkeays
```
rpetry: No such thing as excessive documentation - yours is just fine..;)
This comment is probably an exception though - as the git commit message is dated and says the same. Probably worth removing.

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+# 	- filename <- the name of the NetCDF file
+#	- accnum <- this is given a default value of NULL
+#
+# These comments --> ?
+# Post-AEW migration TODOs:
+# * NChannelSet
+# * Multiple array designs
+#
+# <<- means global assignment
+read.atlas.nc <<- function (filename, accnum = NULL) {
```
rpetry: A grep through gxa code indicates that this argument is never passed. In an effort to clarify the code it might be worth removing it from the argument list.

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
-    rownames(bdc) = de
-    colnames(bdc) = as
+# read.atlas.nc()
+# 	- Reads data from Atlas NetCDF file and builds a Biobase ExpressionSet
+# 	object from it.
+# ARGUMENTS:
+# 	- filename <- the name of the NetCDF file
+#	- accnum <- this is given a default value of NULL
+#
+# These comments --> ?
```
rpetry: Let's remove these comments - we cam only posit what they mean and they are now out of date.

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+# Post-AEW migration TODOs:
+# * NChannelSet
+# * Multiple array designs
+#
+# <<- means global assignment
+read.atlas.nc <<- function (filename, accnum = NULL) {	
+    
+	# Load some packages -- why require() and not library()?
```
rpetry: Yeah, according to http://stackoverflow.com/questions/5595512/what-is-the-difference-between-require-and-library:
'require is used inside functions, as it outputs a warning and continues if the package is not found, whereas library will throw an error'
But I don't quite see the point of continuing if these libraries are not present, as we're bound to get an error eventually. Give the above, I'd change all require() statements to library()

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	# 	- ncdf for reading NetCDFs
+	#	- Biobase so we can create the ExpressionSet
+	#	- gtools --> ? what's this for
```
rpetry: Would you mind checking http://cran.r-project.org/web/packages/gtools/gtools.pdf and if we don't use any of gtools functions, could we remove. If you do decide to remove it, let's chat about how to test it.

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	require(Biobase)
+	require(gtools)
+
+	# Read data from the NetCDF file and store in netCDF object
+	netCDF = open.ncdf(filename)
+
+
+	# Get some things out of the NetCDF object.
+
+	# as and bs -- these are something like assay names/sample names. Would
```
rpetry: Please do give them better names. You now have our best understanding of this code. If our understanding changes, we can always rename.

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	# like to give them more explicit names.
+	as = as.vector(get.var.ncdf(netCDF, "ASacc"))
+	bs = as.vector(get.var.ncdf(netCDF, "BSacc"))
+
+	# b2a -- this seems to be a mapping between the values in as and bs.
```
rpetry: let's remove 'seems to' from all comments - for the reason I mentioned above.

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	# The fixMatrix() function is defined elsewhere in this file.
+	b2a = fixMatrix(get.var.ncdf(netCDF, "BS2AS"), nRows = length(as), nCols = length(bs))
+
+	# designElements -- renamed (used to be called 'de'); this is a vector of
```
rpetry: "designElements -- renamed (used to be called 'de');" - this can go

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	# the design element (a.k.a probe set names)
+	designElements = as.vector(get.var.ncdf(netCDF, "DEacc"))
+
+	# bdc -- want to rename this also. Seems to be the matrix of normalized
```
rpetry: '# bdc -- want to rename this also. Seems to be' - can go also. Feel free to rename it to what makes sense to you.

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	# summarized expressions output by RMA.
+	bdc = fixMatrix(get.var.ncdf(netCDF, "BDC"), nRows = length(as), nCols = length(designElements))
+
+	# See if we've got factors -- these are under "EF" in the NetCDF.
+	if ("EF" %in% names(netCDF$dim)) {
+		
+		# factorNames -- renamed (used to be called 'ef'); this is a vector of
```
rpetry: '# factorNames -- renamed (used to be called 'ef');' can go

rpetry: 'renamed (used to be called 'ef');' can go

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+		# experimental factor names (which in MAGE-TAB is in IDF under
+		# 'Experimental Factor Type' and I think in Atlas under 'property name'
```
rpetry: It's worth discriminating between data and statistics ncdf files. Variables in the data ncdf file are slightly different. In the statistics ncdf file, experimental factor is stored as propertyName.

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+		# (?))
+		factorNames = get.var.ncdf(netCDF, "EF")
+
+		# factorValues -- renamed (used to be called "EFV"); this is a matrix
```
rpetry: '# factorValues -- renamed (used to be called "EFV"); can go

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+		# of factor values.
+		# Each column is an experimental factor; each row is a
+		# sample/assay/scan/hyb <-- what to call it?
```
rpetry: Shall we just leave all four in the comment and remove '<-- what to call it?'

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+		factorValues = data.frame(factorValues)
+
+		# How many factors and assays have we got
+		print(paste("Read in EFV:", nrow(factorValues), "x", ncol(factorValues)))
+	}
+
+	# See if we've got sample characteristics
+	if ("SC" %in% names(netCDF$dim)) {
+		
+		# sampleCharNames -- renamed (used to be called 'sc'); this is a vector
```
rpetry: 'sampleCharNames -- renamed (used to be called 'sc');' can go

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+		# of sample characteristic names (from Characteristics[] in MAGE-TAB).
+		sampleCharNames = get.var.ncdf(netCDF, "SC")
+		
+		# sampleCharValues -- renamed (used to be called 'scv'); this is a
```
rpetry: 'sampleCharValues -- renamed (used to be called 'scv');' can go

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+		
+		# Set row and column names, make it into a dataframe, print how many
+		# characteristics and samples we have
+		colnames(sampleCharValues) = sampleCharNames
+		rownames(sampleCharValues) = bs
+		sampleCharValues = data.frame(sampleCharValues)
+		print(paste("Read in SCV:", nrow(sampleCharValues), "x", ncol(sampleCharValues)))
+	}
+	
+	# this line was commented out
```
rpetry: Let's remove it then - it'll still be preserved inside previous version of this file in git

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	# deacc = get.var.ncdf(netCDF, "DEacc")
+	
+
+	# geneNames -- renamed (used to be called 'gn'); this is a vector of gene IDs
```
rpetry: 'geneNames -- renamed (used to be called 'gn'); ' can go

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	# (guessing from Ensembl mappings or something?).
```
rpetry: They are actually internal (numeric) Atlas gene ids; perhaps rename this variable to geneIds ?

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	geneNames = get.var.ncdf(netCDF, "GN")
+
+
+	# this line was commented out
+	# make de's unique
+	#designElements[designElements == 0] <- -(1:length(which(designElements == 0)))
```
rpetry: Let's remove the line if unused

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+
+	# Some experiment information
+	# accnum is experiment accession, e.g. 'E-MEXP-3392'
+	accnum = att.get.ncdf(netCDF,varid = 0,"experiment_accession")$value
+	# qt is quantitation type -- not sure what values go here ('0' in the one I looked at)
+	qt = att.get.ncdf(netCDF,varid = 0,"quantitationType")$value
```
rpetry: I think this code is out of date. ncdump for a data ncdf file shows me global attributes, e.g.
// global attributes:
        :CreateNetCDF_VERSION = "2.0" ;
        :experiment_accession = "E-AFMX-5" ;
        :ADaccession = "A-AFFY-33" ;
Hence I don't think quantitation type is stored in netcdf files any more.
Let's leave it in but mark it as TODO - Review the use of quantitation types here

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	arrayAccession = att.get.ncdf(netCDF,varid = 0,"ADaccession")$value
+	# arrayName -- renamed (used to be called 'adname'); array design name,
+	# perhaps something like 'RaEx-1_0-st-v1' but was '0' in the one I looked
+	# at (E-MEXP-1631_A-AFFY-43).
+	arrayName = att.get.ncdf(netCDF,varid = 0,"ADname")$value
+
+	# Close the NetCDF now we don't need it.
+	close.ncdf(netCDF)
+	
+	# replaceMissingValues() function is defined elsewhere in this file.
```
rpetry: 'replaceMissingValues() function is defined elsewhere in this file.' comment is probably superfluous - the first approach a developer takes is to search for a function in the same file.

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	# Replaces some 'default' values (added in place of missing values in
```
rpetry: Let's change to # Replaces placeholders for missing expression values in NetCDF with 'NA'

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	# accnum is experiment accession, e.g. 'E-MEXP-3392'
+	accnum = att.get.ncdf(netCDF,varid = 0,"experiment_accession")$value
+	# qt is quantitation type -- not sure what values go here ('0' in the one I looked at)
+	qt = att.get.ncdf(netCDF,varid = 0,"quantitationType")$value
+	# arrayAccession -- renamed (used to be called 'adacc'); array design accession, e.g. 'A-AFFY-99'
+	arrayAccession = att.get.ncdf(netCDF,varid = 0,"ADaccession")$value
+	# arrayName -- renamed (used to be called 'adname'); array design name,
+	# perhaps something like 'RaEx-1_0-st-v1' but was '0' in the one I looked
+	# at (E-MEXP-1631_A-AFFY-43).
+	arrayName = att.get.ncdf(netCDF,varid = 0,"ADname")$value
```
rpetry: Let's add a TODO here also - as that's not stored in data ncdf file either

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	# Get some things out of the NetCDF object.
+
+	# as and bs -- these are something like assay names/sample names. Would
+	# like to give them more explicit names.
+	as = as.vector(get.var.ncdf(netCDF, "ASacc"))
+	bs = as.vector(get.var.ncdf(netCDF, "BSacc"))
+
+	# b2a -- this seems to be a mapping between the values in as and bs.
+	# The fixMatrix() function is defined elsewhere in this file.
+	b2a = fixMatrix(get.var.ncdf(netCDF, "BS2AS"), nRows = length(as), nCols = length(bs))
```
rpetry: Feel free to rename b2a to something more descriptive

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	}
+
+	# Set row and column names for the normalized data matrix and 'b2a'.
+	rownames(bdc) = designElements
+	colnames(bdc) = as
+
+	rownames(b2a) = as
+	colnames(b2a) = bs
+	
+	# Say which experiment we're looking at and how big the normalized data
```
rpetry: Let's use log instead of say or show. thanks

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	# matrix is.
+	print(paste("Read in", accnum))
+	print(paste("Read in BDC:", nrow(bdc), "x", ncol(bdc)))
+
+	# Get the experiment and array design accessions from the NetCDF filename.
+	ncinfo = unlist(strsplit(basename(filename),"_|[.]"))
+	exptid = ncinfo[1]
```
rpetry: General assumption in programming is that ids are numeric. The ncdf file names used to have numeric ids in them - now they have accession. COuld you rename both vars to say expAcc and arrayDesignAcc ?

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	arraydesignid = ncinfo[2]
+
+	# Create factorsForEset data frame (renamed -- used to be 'efscv'). If
```
rpetry: (renamed -- used to be 'efscv') - can go

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	# make de's unique
+	#designElements[designElements == 0] <- -(1:length(which(designElements == 0)))
+
+	# Some experiment information
+	# accnum is experiment accession, e.g. 'E-MEXP-3392'
+	accnum = att.get.ncdf(netCDF,varid = 0,"experiment_accession")$value
+	# qt is quantitation type -- not sure what values go here ('0' in the one I looked at)
+	qt = att.get.ncdf(netCDF,varid = 0,"quantitationType")$value
+	# arrayAccession -- renamed (used to be called 'adacc'); array design accession, e.g. 'A-AFFY-99'
+	arrayAccession = att.get.ncdf(netCDF,varid = 0,"ADaccession")$value
```
nsklyar: I'd keep arrayDesignAcc, since everywhere in Atlas code we call it "array design"

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+		factorsForEset <- data.frame(row.names=as)
+	}
+	
+
+	# Create some Biobase AnnotatedDataFrame objects for addition to the ExpressionSet.
+	
+	# fDataFrame is a mapping between the gene IDs and design element (probe set) names.
+	# Have changed 'gn' and 'de' to 'geneNames' and 'designElements' everywhere
+	# but in the column heading names here -- not sure but afraid it might break
+	# things if I change these. Though actually, can't see that these headings
```
rpetry: I agree - I don't think they are used either - let's rename to what we feel is sensible. If it breaks, at least we will understand the code better. We will need to test the whole thing before merging into our main (develop) branch anyway.

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	fDataFrame = data.frame(gn = geneNames, de = designElements) #, deacc = deacc)
+	fData = new("AnnotatedDataFrame", data = fDataFrame)
+	# Set feature names to the design element names
+	featureNames(fData) = designElements
+
+	# pData contains the factor info.
+	pData = new("AnnotatedDataFrame", data = factorsForEset)
+
+
+	# If we've got sample characteristics, make one for them too.
```
rpetry: If it's not used, let's remove this statement

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+		other = list(accession = accnum,
+			experimentid = exptid,
+			arraydesignid = arraydesignid,
+			qt = qt,
+			arraydesignaccnum = arrayAccession,
+			arraydesignname = arrayName
+		)
+	)
+
+	# If we've got sample characteristics, add them to the eData object under
```
rpetry: Could we check if this is actually necessary?

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	# attribute 'sampleCharValues' (renamed from 'scv').
+	if (exists("sampleCharValues")) {
+	  attr(eData, "sampleCharValues") <- sampleCharValues
+	}
+
+	# Also add 'b2a' to the eData object
+	attr(eData, "b2a") <- b2a
+
+	# Create an AssayData object containing the normalized summarized
```
rpetry: Could we say # Create an AssayData (cf. limma package) object...

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	# Set the feature names and sample names
+	featureNames(aData) <- designElements
+	sampleNames(aData) <- as
+	
+	# Say what factors we have and how many. Factor names are accessed via
+	# varLabels() function from the pData object.
+	print(paste("Computed phenoData for", paste(varLabels(pData),collapse = ", "), nrow(pData), "x", ncol(pData)))
+	
+	# Create and return a Biobase ExpressionSet object containing the
+	# normalized summarized expressions, factors, design element and gene IDs,
```
rpetry: design element accessions and gene iDs

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	print("Taking log2 of the expression matrix")
+
+	# For most values, 1e-10 is far smaller than the precision expected, so we can add it and be (almost) sure
+	# we've got rid of zeros without disturbing the stats.
+	
+	# Make a new matrix with the log2'ed values in
+	tmp = log2(x + 1e-10)
+
+
+	# Don't totally understand this line. Seem to be replacing infinite values
```
rpetry: I guess the way to look at it is:  
The expression inside the !() brackets is TRUE if
- tmp is is not any of { ‘NA’, ‘NaN’, ‘Inf’ or ‘-Inf’ } (c.f. ?is.finite) OR
- tmp is NA
  In such cases we don't want to replace the expr value with 0
  In all other cases we do. 
  Does that make sense?

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
-fstat.eset <-
-  function(eset, varLabel) {
-    print(paste("Calculating design matrix for", varLabel))
-    design = model.matrix(as.formula(paste("~0+", varLabel)), data = eset)
+# fstat.eset() 
+#	- Omnibus one-way ANOVA (with moderated t) F-statistic computation.
+# ARGUMENTS:
+# 	- eset <- an ExpressionSet object containing normalized summarized
+# 	expression data and factor data.
+# 	- factorName <- a factor name (renamed -- used to be called 'varLabel').
```
rpetry: '(renamed -- used to be called 'varLabel').' can go

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+fstat.eset <- function(eset, factorName) {
+    
+	# Say which factor we're looking at.
```
rpetry: Log instead of say

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
     print("Fitting model...")
-    fit = lmFit(eset, design)
 
+	# Run lmFit() on the data in eset using the design matrix.
+	# lmFit() fits a linear model for each probe set on the array:
+	#
+	# 	Y_{i} = b_{0} + b_{1}X_{i} + e_{i}
+	# 
+	# Where Y_{i} is expression from arrays with one factor value and X_{i} is
+	# expression in arrays with the other. Might want more details here?
```
rpetry: Let's remove 'Might want more details here?' and add more info in the future - if necessary

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+	
+	# replaceMissingValues() function is defined elsewhere in this file.
+	# Replaces some 'default' values (added in place of missing values in
+	# NetCDF (?)) with 'NA'.
+	bdc = replaceMissingValues(bdc)
+
+	# If there's only one assay (unlikely?), I think the behaviour is that R
+	# decides 'bdc' and 'b2a' should be vectors instead of matrices. We need
+	# them to be matrices so fix them here.
+	# Otherwise (if >1 assay), just transpose the normalized data matrix.
```
rpetry: Do we know why the matrix needs to be transposed for >1 assay?

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+			numVariableFactorLevels <- nlevels(esetForVariable[[fName, exact=TRUE]])
+
+			# If the number of factor values (levels) is equal to the number of
+			# samples (ncol(exprs(esetForVariable))), this means that there are no
+			# replicates, so we cannot reliably calculate statistics.
+			# If the number of factor values is less than 2 (i.e. every sample is
+			# identical for this factor), it doesn't make sense to look for
+			# differential expression.
+			#
+			# Wondering about this bit. In Kapushesky et al. 2010
```
rpetry: This is a _very_ good point! This logic may have never been correct... Do you think you could change the logic to check each efv for at least two replicates? I also agree with your suggestion to follow Benilton's advice - we suffer from far too many false positives if anything. Shall we go for min replicates - middle of the road? We'll do an exhaustive test - recalculating statistics for all experiments in Atlas and then run the sanity test to see how many end up with no statistics at all. If the number is minimal, we're good to go (we can perhaps consider unloading such experiments). Otherwise we'll need to see - either way it will be an extremely useful finding about our data - thanks for doing this analysis!

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+            #            [,1]       [,2]       [,3]
+			# [1,]  0.6666667 -0.3333333 -0.3333333
+			# [2,] -0.3333333  0.6666667 -0.3333333
+			# [3,] -0.3333333 -0.3333333  0.6666667
+			# 
+			# So with this matrix we are asking e.g. for row 1: is the
+			# expression of samples with factor value 1 significantly different
+			# from the average expression of samples with all other factor
+			# values? Etc. (I think, maybe we can check/discuss this...).
+			# Renaming contrastsMatrix (used to be called 'cm').
```
rpetry: ' # Renaming contrastsMatrix (used to be called 'cm').' can go

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+			#
+            #            [,1]       [,2]       [,3]
+			# [1,]  0.6666667 -0.3333333 -0.3333333
+			# [2,] -0.3333333  0.6666667 -0.3333333
+			# [3,] -0.3333333 -0.3333333  0.6666667
+			# 
+			# So with this matrix we are asking e.g. for row 1: is the
+			# expression of samples with factor value 1 significantly different
+			# from the average expression of samples with all other factor
+			# values? Etc. (I think, maybe we can check/discuss this...).
```
rpetry: Let's remove '(I think, maybe we can check/discuss this...)' - if we find out otherwise, we'll amend the documentation as necessary

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+			# Calculate moderated t-statistics again using eBayes().
+			# This is a robust version of the t-test which reduces the false discovery
+			# rate. Standard errors are moderated across genes (shrunk towards a common
+			# value). A gene that has a large variance is less likely to actually be
+			# differentially expressed.
+			contr.fit = eBayes(contr.fit)
+			
+			# Use decideTests() to call each gene significantly differentially
+			# expressed (or not). Based on t-statistics and p-values, it
+			# decides whether to call a gene 'up', 'down', or not significant.
```
rpetry: Change 'not significant' to 'non-differentially expressed'?

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
+			# method="global" treats the entire matrix of t-statistics as a
+			# single vector of unrelated tests. We use default p-value of 0.05.
+			#
+			# Adjusted p-values are calculated using Benjamini and Hochberg
+			# (1995) method (fdr) to correct for multiple testing by
+			# controlling the false discovery rate.
+			#
+			# Where do the results of this actually get used?
```
rpetry: Yeah, good point. On the face of it, it looks as if only colnames(dec) are only ever used? We could do with some more research on this - and remove it if not used. Otherwise calling decideTests just clouds the issue. Perhaps in the past the differential expression call was meant o come from with R, whereas now we make that decision outside of R. 
We'll just need to think of how to pass colnames to tab$t and tab$p.value.adj in computeAnalytics without having to call decideTests()

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -1,211 +1,520 @@
...
-        allFits[[varLabel]] = thisFit
-        print("Done.")
-      })
-    }
-
-    allFits
-}
 
 ### Atlas analytics processing driver: read the data, compute the linear fit, post-hoc test, adjust and write to tab-delim files
 process.atlas.nc<-
```
rpetry: I can't see anywhere this function is called in Atlas. Have you ever found it useful in your investigations?

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+		# expressions and experimental data such as factors and their values.
+		# Created using read.atlas.nc() function defined elsewhere in this
+		# file.
+		eset = read.atlas.nc(data_ncdf)
+		
+		# data_nc is an ncdf object with experiment data.
+		data_nc = open.ncdf(data_ncdf)
+		# statistics_nc is an ncdf object opened for writing; so if there is
+		# already a stats NetCDF, things are overwritten when this is run. What
+		# happens for a new experiment? There has to be a file with something
```
rpetry: A statistics ncdf file is not created when an experiment is (re-)loaded, but created just before analytics are calculated. Perhaps you could amend the comment in the light of this.

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+
+		# Stick them together with paste :)
+		# pairs is a vector with strings "factor name||factor value" for every
+		# factor name/value pair in the experiment.
+		pairs <- paste(propertyNAME, propertyVALUE, sep = "||")
+
+		# Create two matrices called tstat and pval and fill with 'NA's. The
+		# number of rows is equal to the number of design elements/probe sets;
+		# the number of columns is equal to the number of factor name/value pairs.
+		tstat = matrix(NA, ncol = length(propertyNAME), nrow = nrow(eset)); #t(get.var.ncdf(statistics_nc, "TSTAT"))
```
rpetry: Do we need these transpose comments at the end?

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+				# Another copy of the p-values
+				pv = tab$p.value
+				# Which ones aren't 'NA'
+				o = !is.na(tab$p.value)
+				# Run Benjamini and Hochberg fdr correction on the non-NA p-values.
+				pv[o] = p.adjust(pv[o], method = "fdr")
+				# Add the adjusted p-values to tab
+				tab$p.value.adj = pv
+				
+				# Put the calls from decideTests() (-1 = 'down'; 0 =
```
rpetry: Please review this in light of my previous comments about decideTests()

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+				# non-significant; 1 = 'up') in tab under tab$Res. What happens
+				# next with these?
+				tab$Res <- unclass(proc[[factorName, exact = TRUE]]$boolupdn)
+				
+
+				# The following 3 lines were commented out.
```
rpetry: Let's remove commented out code

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+				# tab$F <- proc[[factorName, exact = TRUE]]$F
+				# tab$F.p.value <- proc[[factorName, exact = TRUE]]$F.p.value
+				# tab$F.p.value.adj = proc[[factorName, exact = TRUE]]$F.p.value.adj
+
+
+				# I think that 'genes' gets the design element names? Put these in tab.
```
rpetry: hmm, I can't quite see where $genes comes from or what it contains.. Is tab$Genes actually used anywhere?

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+				# tab$p.value.adj.
+				pval[,which(colnames(pval) %in% colnames(tab$p.value.adj))] <<- tab$p.value.adj[,colnames(pval)[which(colnames(pval) %in% colnames(tab$p.value.adj))]]
+				
+				# Now we've stored the info we need, return "OK" for this factor.
+				return("OK")
+			}
+		)
+
+		
+		# Put the contents (transposed) of tstat and pval  the statistics
```
rpetry: Do we know why transposition is necessary?

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+		# ncdf object.
+		print(paste("Writing tstat and pval to NetCDF:", ncol(tstat), "x", nrow(tstat)))
+		put.var.ncdf(statistics_nc, "TSTAT", t(tstat))
+		put.var.ncdf(statistics_nc, "PVAL", t(pval))
+		
+		# Get the factor names from the NetCDF.
+		efsc = get.var.ncdf(data_nc, "EF")
```
rpetry: Can we rename efsc to ef? efsc comes from the time where we planned to calculate stats for sample characteristics values (as well as factor values). Since then we thought better..

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+	# property name entries in the NetCDF, which is equal to the total number
+	# of factor name/value pairs (e.g. 'compound:none', 'compound:NaCl',
+	# 'genotype:wild type', 'genotype:mutant').
+	nCols <- length(get.var.ncdf(statistics_nc, "propertyNAME"))
+	# Read in the t-statistics and p-values and transpose with
+	# transposeMatrix() function (defined elsewhere in this file).
+	pval <- transposeMatrix(get.var.ncdf(statistics_nc, "PVAL"), nCols)
+	tstat <- transposeMatrix(get.var.ncdf(statistics_nc, "TSTAT"), nCols)
+	
+	# Get the gene names (renamed to geneNames from 'gn').
```
rpetry: '(renamed to geneNames from 'gn') can go. Could we call it geneIds pls?

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+	geneNames <- get.var.ncdf(data_nc, "GN")
+
+	# Say how much data we've just read in. Number of rows is equal to number
+	# of design elements; number of columns is equal to number of factor
+	# name/value pairs.
+	print(paste("T(rows:", nrow(tstat), "cols:", ncol(tstat), ")"))
+	print(paste("P(rows:", nrow(pval), "cols:", ncol(pval), ")"))
+
+	# Replace some 'default' values added in place of missing values using
```
rpetry: Could we change the comment according to my previous suggestion

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+	# replaceMissingValues() function (defined elsewhere in this file).
+	tstat <- replaceMissingValues(tstat)
+	pval <- replaceMissingValues(pval)
+
+	# Find row indices of gene names that are 0. I guess we don't want to show
+	# info for design elements not mapped to genes?
+	zeroGnIdxs <- (geneNames == 0)
```
rpetry: yes, geneId == 0 corresponds to a 'no gene'

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+	# naIdxsT and naIdxsP are vectors of TRUEs and FALSEs for each row in tstat
+	# or pval (TRUE if that row has all NA; FALSE if not).
+	naIdxsT <- apply(is.na(tstat), 1, all)
+	naIdxsP <- apply(is.na(pval), 1, all)
+	
+	# naIdxs is a vector of TRUEs and FALSEs for each row in tstat and pval
+	# where either tstat or pval (or both) has all NA.
+	naIdxs <- apply(cbind(naIdxsT, naIdxsP), 1, function(x){ x[1] || x[2]})
+	
+	# Say how many NA rows we found in pval.
```
rpetry: Log

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+	print(paste("length( NA rows ):", length(which(naIdxsP))))
+
+	# allBadIdxs is a vector with TRUE if pval or tstat was NA or the
+	# corresponding gene name was 0 for that row.
+	allBadIdxs <- apply(cbind(naIdxs, zeroGnIdxs), 1, function(x){ x[1] || x[2] })
+
+	# Now rank the genes based on each of five conditions ('statfilter's):
+	# 	- "ANY": the top significantly differentially expressed genes,
+	# 	regardless of 'up' or 'down' call, as well as the non-significantly
```
rpetry: could we shorten 'non-significantly differentially expressed' to 'non-differentially expressed' ?

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+	# 	differentially expressed genes.
+	# 	- "UP_DOWN": the top significantly expressed genes, regardless of 'up'
+	# 	or 'down' call.
+	# 	- "UP": the top significantly up-regulated genes.
+	# 	- "DOWN": the top significantly down-regulated genes.
```
rpetry: ditto

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+	# 	- "NON-DE": the top non-significantly differentially expressed genes.
```
rpetry: ditto

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
+		ifelse (statfilter == "ANY",
+			badIdxs <- zeroGnIdxs,
+			badIdxs <- allBadIdxs
+		)
+
+		# Subset only the rows of tstat and pval that do not have indices matching ones in badIdxs.
+		tstatGood <- filterMatrix(tstat, 1, !badIdxs)
+		pvalGood <- filterMatrix(pval, 1, !badIdxs)
+
+		# Say what we're doing now.
```
rpetry: Log

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
-                         tstat[,which(colnames(tstat) %in% colnames(tab$t))] <<- tab$t[,colnames(tstat)[which(colnames(tstat) %in% colnames(tab$t))]]
-                         pval[,which(colnames(pval) %in% colnames(tab$p.value.adj))] <<- tab$p.value.adj[,colnames(pval)[which(colnames(pval) %in% colnames(tab$p.value.adj))]]
+# replaceMissingValues()
+# 	- Replaces certain values in a matrix with NA. Assume these are default
+# 	missing values for NetCDFs?
+# ARGUMENTS:
+# 	- m <- a matrix (of either normalized summarized expressions or
+# 	t-statistics or p-values).
+replaceMissingValues <<- function(m) {
+	m[m <= -1e6] = NA
```
rpetry: Comment on the use of -1e6 value by Atlas at experiment load time (to represent non-float expression values in the loaded data files)

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
 
-fixMatrix <<-
-  function(m, nCols, nRows) {
-     if (is.matrix(m)) {
-       return(m)
-     }
-     ifelse(nCols > 0, out <- matrix(m, ncol = nCols), out <- matrix(m, nrow = nRows))
-     return(out)
-  }
+# This function is not called anywhere in this file.
```
rpetry: Let's remove it

atlas-analytics/src/main/resources/R/analytics.R
```diff
@@ -236,249 +545,603 @@ process.atlas.nc<-
...
-    }
 
-    for (i in seq_along(maxtstatidxs)) {
-      minpvals[i] <- f.pval[i, maxtstatidxs[i]]
-      maxtstats[i] <- f.tstat[i, maxtstatidxs[i]]
-    }
+# orderByStatfilter()
+#	- Sorts t-statistcs and p-values according to any of 5 'statfilter's:
+#	 	- "ANY": the top significantly differentially expressed genes,
+#	 	regardless of 'up' or 'down' call, as well as the non-significantly
```
rpetry: 'non-significantly differentially expressed' to 'non-differentially expressed' everywhere - thanks.

atlas-loader/src/main/resources/R/normalizeOneExperiment.R
```diff
@@ -1,70 +1,171 @@
...
+		write(c("Scan REF", scansSorted), outFile, ncolumns = len+1, sep = "\t")
+		
+		# Not sure why we need this row. Repeats "GEO:AFFYMETRIX_VALUE" for
+		# each column. Don't know what happens to this textfile between its
+		# creation here and analytics.R -- maybe it's needed then or something
+		# (or does analytics.R expect it? can't remember).
+		write(c("Composite Element REF", rep("GEO:AFFYMETRIX_VALUE", len)), outFile, ncolumns = len+1, sep="\t", append=TRUE)
+
+
+		# This line was commented out.
```
rpetry: I'd remove it - as it is not used and we don't know why it's there

atlas-loader/src/main/resources/R/normalizeOneExperiment.R
```diff
@@ -1,70 +1,171 @@
...
+		len <- length(sampleNames(eSet))
+		
+		# The scan names after sorting
+		write(c("Scan REF", scansSorted), outFile, ncolumns = len+1, sep = "\t")
+		
+		# Not sure why we need this row. Repeats "GEO:AFFYMETRIX_VALUE" for
+		# each column. Don't know what happens to this textfile between its
+		# creation here and analytics.R -- maybe it's needed then or something
+		# (or does analytics.R expect it? can't remember).
+		write(c("Composite Element REF", rep("GEO:AFFYMETRIX_VALUE", len)), outFile, ncolumns = len+1, sep="\t", append=TRUE)
```
rpetry: Yeah, am not sure - the normalized file is read in by readDataMatrixFile() in  in atlas-loader/src/main/java/uk/ac/ebi/gxa/loader/datamatrix/DataMatrixFileBuffer.java. I'll experiment with the normalized file looks like and whether not having the "Composite Element REF" line makes a difference.

atlas-analytics/src/main/resources/R/analytics.R
mkeays: It doesn't look like any of those functions are used -- commenting out library(gtools) .

atlas-analytics/src/main/resources/R/analytics.R
mkeays: re b2a -- I actually can't see where it is used after its addition to the ExpressionSet. Any ideas why it might be needed?

atlas-analytics/src/main/resources/R/analytics.R
mkeays: Yeah, I don't think that it is, at least for the usual computing analytics process. The only place I can see the info in eData being accessed is in the function process.atlas.nc , which is for writing to tab-delimited text files rather than to NetCDFs, and we don't know if that function is ever used anywhere. I've never used it.

Therefore, tempted to remove eData completely, plus the `if ("SC" %in% names(netCDF$dim))` statement (line 84) and other references to Characteristics.

atlas-analytics/src/main/resources/R/analytics.R
mkeays: Renaming them to ncExptAcc and ncArrayDesignAcc to distinguish them from other vars with experiment and array design accessions as ones derived from the NetCDF filename. Not sure we need them though, they go into experimentData in the ExpressionSet which is only accessed by process.atlas.nc().

atlas-analytics/src/main/resources/R/analytics.R
mkeays: I think so... I think I was expecting that it would get rid of all NAs,  but in fact we keep the ones that were NA when we started and only replace with 0 NAs (and infinite values etc) that were created during the log2() ?

atlas-analytics/src/main/resources/R/analytics.R
mkeays: OK, can you remember/do you have more notes from Benilton's class? I just can't remember if this is right, and I get confused by talk of "comparing expression against the average for _all_ samples" -- this isn't what I'm saying in this comment. Actually is that what the contrasts matrix does?

atlas-analytics/src/main/resources/R/analytics.R
mkeays: No, I sort of assumed it was something to do with NetCDFs, because we get the matrix out of the NetCDF and transpose, and then when we put it back into the NetCDF at the end we transpose it back.

atlas-analytics/src/main/resources/R/analytics.R
mkeays: Yes can have a go. When you say "Shall we go for min replicates - middle of the road?" are we talking min 3 replicates?

atlas-analytics/src/main/resources/R/analytics.R
mkeays: Yeah, `colnames(dec) = levels(esetForVariable[[fName, exact = TRUE]])` is just setting the column names of dec. dec is put into thisFit under $boolupdn. It is copied from here to tab$Res in computeAnalytics(), but as far as I can see not used after that. It is also accessed within process.atlas.nc(), and written to the tab-delimited file it produces.

Coming back to this later I see what you mean now actually... but think that it shouldn't be too hard to just pass the column names so they can be matched for tstat and pval.

atlas-analytics/src/main/resources/R/analytics.R
mkeays: Nope never used it, didn't know about it til last week :)

atlas-analytics/src/main/resources/R/analytics.R
mkeays: Yeah don't think tab$Res is used again. I wonder if historically process.atlas.nc() came first, and things were copied to make computeAnalytics(), and now some stuff is non-functional -- pseudogenisation? ;) 

The tab object is also created in process.atlas.nc() and this is all written out to the tab-delim file.

atlas-analytics/src/main/resources/R/analytics.R
mkeays: No, only in process.atlas.nc() where tab is also created and written to tab delim file. Actually I think that we could get rid of tab completely from computeAnalytics().

atlas-analytics/src/main/resources/R/analytics.R
mkeays: No, assumed it was something to do with the NetCDF since it's transposed to be put in there and transposed back again when we get it out.

atlas-loader/src/main/resources/R/normalizeOneExperiment.R
mkeays: OK thanks. I can now say for certain it's not needed in analytics.R :)

atlas-analytics/src/main/resources/R/analytics.R
rpetry: If anything, it may have been added early on just in case it was needed? I agree it is not used anywhere - let's remove it pls.

atlas-analytics/src/main/resources/R/analytics.R
rpetry: Yes, let's remove both what you suggest and process.atlas.nc. If we ever need a version to write to tab-delimited format, we will re-add it. The stats code will no doubt change in the next few months and if we leave process.atlas.nc in, it'll just get more and more out of date. Thanks.

atlas-analytics/src/main/resources/R/analytics.R
rpetry: Since we're now dropping process.atlas.nc, let's drop them too.

atlas-analytics/src/main/resources/R/analytics.R
rpetry: I suspect the fact that is.finite() includes a check for NA is incidental to the logic intended behind 'tmp[!(is.finite(tmp) | is.na(x))] <- 0'. I think it is meant to replace with 0 just minus infinite values (resulting from log2(~0), and leave all NA's alone.

atlas-analytics/src/main/resources/R/analytics.R
rpetry: Would you mind looking at an example ncdf data file (perhaps using R ncdf library) to confirm why we transpose the matrix for >1 assay, but apparently not just for 1 assay (though I assume that matrix(bdc, nrow = length(designElements) achieves the same) ?

atlas-analytics/src/main/resources/R/analytics.R
rpetry: Yeah, sorry, I meant 3 replicates. As we are also replacing 'at least one efv with min X replicates' to 'each efv with min X replicates', I don't want to end up being over-restrictive. Let's re-calculate stats for all experiments with min 3 replicates per efv and see how much difference that makes.

atlas-analytics/src/main/resources/R/analytics.R
rpetry: Great - let's try and do that then (as we're now also dropping process.atlas.nc)

atlas-analytics/src/main/resources/R/analytics.R
rpetry: Yeah, let's drop it for the reasons mentioned earlier.

atlas-analytics/src/main/resources/R/analytics.R
rpetry: Yes, let's eliminate decideTests()

atlas-analytics/src/main/resources/R/analytics.R
rpetry: yes let's get rid of tab.

atlas-analytics/src/main/resources/R/analytics.R
rpetry: let's check in ncdf and put a comment in on why we need to transpose.

atlas-analytics/src/main/resources/R/analytics.R
mkeays: Sure will take a look. You're right about matrix(bdc, nrow=length(designElements)) though I think: for 1 assay, R creates a vector, and you can create a matrix of any dimensions you want just by specifying nrow or ncol (or both). Since the number of elements in the vector for one assay is the same as the number of design elements, using nrow=length(designElements) means it's a matrix with only one column. Will amend the comment to reflect that and look into ncdf.

atlas-analytics/src/main/resources/R/analytics.R
rpetry: Yes, I'm not yet clear on this. I assume you have access to http://www.ebi.ac.uk/industry/private/Documents/workshop-materials/120530_Rbio/RBioC-WorkshopPresentation.pdf. Feel fre to put in the documentation you are confident with and leave out the detail you're not sure of. I'll need some time to get to the stage I feel I understand this well enough to comment authoritatively. Thanks.

atlas-loader/src/main/resources/R/normalizeOneExperiment.R
rpetry: I've checked in DataMatrixFileBuffer.java that all header lines (including "Composite Element REF" ) are ignored when data is read in. Hence we should be OK removing this line.

atlas-analytics/src/main/resources/R/analytics.R
mkeays: Thanks, I forgot I had that. Will re-read...

atlas-analytics/src/main/resources/R/analytics.R
mkeays: So in the ncdf the normalized data is stored as: `"float BDC[AS,DE]  Longname:BDC Missval:1e+30"`. I don't know enough about creating ncdfs (could keep looking into it...) but it looks like the matrix is "BDC", with dimensions "AS" rows by "DE" columns where "AS" is number of assays and "DE" is number of design elements. It seems that if you try to write the data matrix back to the ncdf untransposed, it still ends up with these dimensions, not sure how, i.e. whether it transposes it or does something else...

I haven't come across anything saying that you can't store matrices with large numbers of rows (e.g. "DE" in the one I looked at was 31099). I guess that we can't really change it without going to change every single ncdf in Atlas though. I can add to the comments that we transpose because of the way the data is stored in the ncdfs.

atlas-analytics/src/main/resources/R/analytics.R
mkeays: Removing qt as well as the array design accession and name -- these aren't used anywhere other than process.atlas.nc and we are removing that. The only one from this list to keep is accnum since it's used in log further down.

atlas-analytics/src/main/resources/R/analytics.R
mkeays: OK understand this much better now :) updated the comments for those parts.

atlas-analytics/src/main/resources/R/analytics.R
mkeays: Testing factor replicates stuff on local machine with library(gtools) commented out, get error message:
`Error in as.character(con) : 'con' is missing`
Does not happen with library(gtools) uncommented -- so we DO need it! Not totally sure for which bit though. Will investigate it later...

atlas-analytics/src/main/resources/R/analytics.R
mkeays: OK cool, now checking for >= 3 replicates for every factor value (i.e. skipping if one factor value has <3 reps).

atlas-analytics/src/main/resources/R/analytics.R
mkeays: Embarrassingly, this isn't happening any more and things are running happily on my laptop without gtools... I don't know what the above message was about! Can't reproduce it.

PR #339, opened 2012-06-22
added export of  organisms from ensembl

nsklyar:


nsklyar:
data are available on [gxa]/dataExport/organisms

PR #340, opened 2012-07-11
Trac3525 - Basic experiment search functionality

rpetry:


atlas-web/src/main/java/ae3/dao/DAOException.java
```diff
@@ -0,0 +1,45 @@
...
+
+package ae3.dao;
+
+/**
+ * User: nsklyar
+ * Date: 29/06/2012
+ */
+public class DAOException extends RuntimeException{
+    public DAOException() {
+        super();    //To change body of overridden methods use File | Settings | File Templates.
```
rpetry: Do we need these default comments in the code?

